PR #65, opened 2022-02-11
Correct port reference in README

JamesMGreene:
Per the code, it looks like we default to port `8080` rather than the mentioned `8081`. 🧐


JamesMGreene:
@waldnzwrld I don't follow where this port `8081` you're suggesting is coming from, though...? 🤔 

Our new server code for loading the frontend via Express defaults to `8080` is not specified: 
https://github.com/ProgramEquity/amplify/blob/96850125e2a6107ea9de436ffcfd05ed35d75bc4/server/server.js#L2

Also, if using the `vue-cli-service serve` command instead, that also defaults to `8080` [per their docs](https://cli.vuejs.org/guide/cli-service.html#vue-cli-service-serve):

> ```
> Usage: vue-cli-service serve [options] [entry]
> 
>   ...
>   --port         specify port (default: 8080)
> ```

JamesMGreene:
@waldnzwrld I'm going to merge this for now. 😅 

That said, if I'm off-base here after all, please let me know so we can make these values consistent going forward! 🙏🏻 🙇🏻‍♂️

PR #70, opened 2022-02-16
Add nodenv to brewfile, install node through nodenv

waldnzwrld:
This PR adds nodenv to the brewfile and then uses nodenv to install the appropriate node version for this project


JamesMGreene (approved):
LGTM! 🚀 

This should also hopefully fix issues folks have been running into where they've installed Node.js globally and then the `brew bundle` fails. 🤞🏻

PR #72, opened 2022-02-16
Handle differentiated deployments for frontend vs. backend

JamesMGreene:
Prior to combining these two repos, our deployments worked as follows:
- For deploying our frontend to Heroku, we relied on Heroku invoking the `npm run build` command via the Node.js buildpack to generate the `dist/` directory of transpiled files.
- For deploying our backend to Heroku, we relied on Heroku invoking the `npm run heroku-postbuild` command via the Node.js buildpack to run the production database migrations.

However, now that the codebases are combined, both Heroku deployments would always be invoking `npm run heroku-postbuild` instead as [the Node.js buildpack will prefer that script](https://devcenter.heroku.com/articles/nodejs-support#customizing-the-build-process), if defined.

As such, deploying the frontend from the current codebase would both:
- (a) not build the `dist/` directory, and
- (b) probably outright **_fail_** because the `DATABASE_URL` wouldn't be defined when Knex tries to invoke the database migrations.

That's not good. 🙃 

So this PR introduces a new script `script/heroku-postbuild.sh` that will get invoked by both deployments instead, and then does some naïve environment variable analysis to see which deployment is being attempted in order to execute the actually desirable commands. 😅 

**TODO after merging:**
- [ ] Change the repository reference in _ALL_ the Heroku Apps (backend + various frontends) and try a manual deployment
- [ ] For each App, if the manual deployment works, enable the Automatic Deployment feature


JamesMGreene:
More to be done here, hang tight. 👍🏻

JamesMGreene:
With these changes in place, I believe we could once again deploy the frontend and backend to _two separate Heroku apps_.

ℹ️ NOTE: That is _not_ the final state I would like to see, but it is a stepping stone on getting this new combined codebase back into a healthier state first. 👌🏻 

I'll try testing this in 2 isolated Heroku apps first to verify before merging. 🧪

JamesMGreene:
Successfully deploy the frontend to an isolated environment using this PR, though I'll probably tear it down right away to save Heroku dollars. 💰 🔥 

Was at: https://amplify-combined-frontend.herokuapp.com/

Kapture 2022-02-17 at 21 43 59

JamesMGreene:
Successfully deploy the backend to an isolated environment using this PR, though I've already torn it down right away to save Heroku dollars. 💰 🔥 

All I needed to do was:
1. Create the Heroku App
2. Configure the Heroku Buildpack (Node.js)
3. Add all of the necessary secrets as Config Vars in the App
4. Connect the App to this repository on this PR's branch

Was at: https://amplify-combined-backend.herokuapp.com/

Kapture 2022-02-17 at 22 48 28

manishapriya94:
@bergsm this is a good PR to review to get up to speed!

JamesMGreene:
Closing this out in favor of PR #87

PR #76, opened 2022-02-18
Tighten the version of Node that is valid

JamesMGreene:
Our Heroku deployments were complaining about our potentially dangerous version specification for the acceptable Node.js engines as it is allowing for experimental versions rather than only stable/LTS versions.

image


JamesMGreene:
This is causing deployment failures so I'm going to merge it now. 😅

Issue #49, opened 2021-09-07
Review Letter: Customize Letter


manishapriya94:
### Screenshot
<img width="455" alt="Screen Shot 2022-02-10 at 8 36 46 AM" src="https://user-images.githubusercontent.com/66452376/153441710-9e71c674-df91-495b-94b1-e99b2554c2e2.png">

Advocacy groups singly pointed out that customized letters were the most significant way for public leaders to resonate with their constituents' perspective - its a way for us to paint what the world will look like with or without this request for change. 

### Tasks
- [ ] Understand how [Lob templates ](https://docs.lob.com/#operation/template_retrieve) are currently built in our [api](https://github.com/ProgramEquity/amplify-back-end/wiki/Lob-API-Endpoints)
- [ ] Figure out if there is a way to append letters with a customization textbox without having to create a new template as mentioned here in the [Lob documentation](https://docs.lob.com/#operation/template_retrieve) in the screenshot below
<img width="711" alt="Screen Shot 2022-02-10 at 8 26 27 AM" src="https://user-images.githubusercontent.com/66452376/153439742-b6197df8-fd2a-410e-b5f6-b1dc74615c5f.png">

PR #81, opened 2022-02-22
Add static customize input field for letter

JamesMGreene:
As mentioned in https://github.com/ProgramEquity/amplify/issues/49

Recreating https://github.com/ProgramEquity/amplify-front-end/pull/101

cc @aneelac22

### DIFF
--- a/src/components/LetterLoad.vue
+++ b/src/components/LetterLoad.vue
@@ -20,9 +20,18 @@
         <v-card-title> {{ selectedRep.name }} </v-card-title>
 
         <v-card-text>
-          <span v-html="letterBody"></span>
+          <span class="text-sm-left" v-html="letterBody"></span>
         </v-card-text>
+
+        <v-text-field
+          label="I'm in support because (This is optional but when you add your story - it makes a difference"
+          :rules="rules"
+          outlined
+          color="blue"
+        >
+        </v-text-field>
       </div>
+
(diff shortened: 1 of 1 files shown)


waldnzwrld (dismissed):

src/components/LetterLoad.vue
```diff
@@ -20,9 +20,18 @@
         <v-card-title> {{ selectedRep.name }} </v-card-title>
 
         <v-card-text>
-          <span v-html="letterBody"></span>
+          <span class="text-sm-left" v-html="letterBody"></span>
         </v-card-text>
+
+        <v-text-field
+          label="I support this cause because... (This is optional but when you add your story - it makes a difference"
```
JamesMGreene, line 27:
```suggestion
          label="I support this cause because... (This is optional but when you add your story - it makes a difference)"
```

JamesMGreene:
@manishapriya94 Could you please confirm if this change is still desirable right now? I recall hearing some discussion about whether this was useful in the short term or not. 🤔

aneelac22:
I think if there was a backend change, I can implement it. But, we shouldn't merge this since, it isn't working

JamesMGreene:
@aneelac22 Thanks for adding that context. ❤️ 🙇🏻

JamesMGreene:
Closing for now for cleanliness. 🧹

PR #85, opened 2022-02-24
Switch back to using Vue's default build location of `dist/`

JamesMGreene:
Vue defaults to building into the `dist/` directory. I can't see a strong reason for why we changed it to go into `server/views/` instead so I'm switching it back for now.

This would also be causing deployment issues as our Heroku frontend app is currently configured to serve files from the `dist/` directory. 😅


waldnzwrld:
The reason this was changed was so that Express could serve the compiled and minified JS and CSS files 

Skipping the minification potentially causes performance issues and exposes raw JS to end users which is a potential security vector

JamesMGreene:
It's still getting built and minified, just into `dist/`. 😊👌🏻

PR #87, opened 2022-02-24
Host the frontend + backend on a single Express port

### LINKED ISSUE #83: Host frontend + backend code on a single port
Given our application's nature, we should ideally just host the frontend + backend code with one Express port (for both dev and production). This should simplify our onboarding for local development, our Codespaces configuration, and our deployment strategy.

In terms of deployment, we would only need a single self-contained Heroku App, and could even consider using other hosting services like Azure, AWS, GCP, etc. 🚀 

While researching this, I found that it is easy enough to have an Express app host a **built** production Vue app but I struggled to find good resources on how to do it in a way that still supported hot-reloading of Vue from a development perspective.

The most interesting resource I found on the topic was [this one](https://dennisreimann.de/articles/vue-cli-serve-express.html), which seemed very promising in its unique approach of mounting the Express app into the Vue app (`vue-cli-service` app, that is) since it already has an internal Express server. 🔄

Related:
- https://github.com/ProgramEquity/amplify-back-end/issues/127

### LINKED ISSUE #84: Use Heroku backend proxying for `/api` to support multiple backends
Alternative to, or stepping stone towards:
- https://github.com/ProgramEquity/amplify/issues/83

---

Currently, the frontend portion of our app contained hardcoded URL references to the current production instance of our backend portion's Heroku App, e.g.:

https://github.com/ProgramEquity/amplify/blob/f8f42e4bf0903694972b6bb81f3bd55682ec34fa/src/components/SearchReps.vue#L107-L109

This is highly restrictive to our continued development as the frontend is always referring to the production backend code, thus not allowing us to develop both the frontend and backend portions of a new feature within a single PR.

This also introduces risk when we discuss having dedicated deployments that will not be updated for special launch events such as:
- https://github.com/ProgramEquity/amplify/issues/79

If we cannot address this with #83 in the short term, one workaround we can use is to use the Heroku static buildpack's [proxy backends feature](https://github.com/heroku/heroku-buildpack-static#proxy-backends).

To implement that would require a few steps.

### Step 1: Remove the explicit domain references

Find all instances of `https://murmuring-headland-63935.herokuapp.com/api/` in the frontend code and replace them with just `/api/`.

:warning: Some extra step probably needs to be taken here to ensure that the local Vue frontend can connect with the local Express backend since they are currently hosted on different ports. Perhaps a Vue Router change or something? I'm not 100% certain on that works, e.g. https://medium.com/bb-tutorials-and-thoughts/vue-js-how-to-proxy-to-backend-server-a562bad965eb

### Step 2: Add a backend proxy configuration to `static.json`

Using a Heroku config var reference like `API_ORIGIN`:

```json
{
  "proxies": {
    "/api/": {
      "origin": "${API_ORIGIN}/api/"
    }
  }
}
```

### Step 3: Create a Heroku config var for the frontend

```
API_ORIGIN=https://murmuring-headland-63935.herokuapp.com
```

### Step 4: Deploy!

Now that everything is configured, we need to ensure our latest code with these changes is deployed to the staging frontend Heroku App! 🚀 

### Step 5: Verify and multiply

After confirming that this works as expected with the staging Heroku App:
- Update the production Heroku App with the config var and redeploy it
- Update the frontend ("hackathon"?) Heroku **pipeline** with the config var

### Step 6: Create more backend environments (if desired)

This will allow us to create a staging environment for the backend code as well, and even a separate PostgreSQL database (probably a good idea to avoid messing up production data) -- if desired! The obvious gotcha to watch out for here would be the Heroku billing costs. 💰 

Just remember to update the frontend config vars to point at your preferred backend URLs per environment. ↔️

JamesMGreene:
Fixes #83
Closes #84 
Closes #72 

Towards:
- #83


JamesMGreene:
Looks like thing are working great when deploying this branch to a new Heroku App! 🚀 

Kapture 2022-02-25 at 12 29 07

PR #90, opened 2022-02-25
Fix outdated repository references

JamesMGreene:
Overlooked fallout from combining the 2 repos for https://github.com/ProgramEquity/amplify-back-end/issues/127


JamesMGreene:
The failures here should be addressed after merging #87

Issue #79, opened 2022-02-21
Sunrise Launch Checklist


manishapriya94:
Thank you folks for coordinating hackathon ships into a recombined repo! Sunrise would like to do a test run on February 26th - is this feasible? 

Heres whats remaining (yellow in [product map](https://docs.google.com/spreadsheets/d/1uEhgbXIrLHz-IrAuiBTRzEMJ_QfahJJlN7J1OmqT-so/edit?usp=sharing)): 

Outreach @manishapriya94 : 
- [ ] add lob letter templates 
- [ ] add each campaign to database 
- [ ] add column in campaigns for logo pictures

Frontend @mahrsbar: 
- [ ] add coupon code ability to payment 
- [ ] finish checkout flow 
- [ ] finish action complete page 

Backend @nawazkhan @teakopp: 
- [ ] update stripe wiki 
- [ ] connect applepay to stripe checkout 

Architecture @JamesMGreene : 
- [ ] create production environment for just sunrise
- [ ] connect it to google domain

PR #92, opened 2022-02-27
Create database migration to add new fields to `campaigns` table

JamesMGreene:
Towards:
- https://github.com/ProgramEquity/amplify/issues/79

cc @manishapriya94 

Adds a database migration to add 3 new **nullable** columns to the `campaigns` table:
- `logo_url` (long text string, or `NULL`)
- `letters_goal` (integer >= 0, or `NULL`)
- `donation_goal` (integer >= 0, or `NULL`)

**Notes:**
- When implementing the frontend views of `letters_goal` and `donation_goal`, it will be important to remember that they can both be `NULL`.
- In the case of `logo_url`, the backend may need to provide an alternative default value.


JamesMGreene:
@manishapriya94 Ready for a quick review when you get a moment. Thanks! 🙏🏻

manishapriya94 (commented):
Looks good, thank you for adding these! @mahrsbar, would you like me to put the logos in our assets into the database for these?

manishapriya94 (approved):

PR #120, opened 2022-03-05
Change zipCode to zip

teakopp:
* Changes zipcode to zip to match var being passed in req.body


JamesMGreene (dismissed):
Should be fine but left an alternative suggestion.

JamesMGreene (approved):

PR #126, opened 2022-03-10
Complete page UI addition

aneelac22:
@manishapriya94 and @mahrsbar I got most of it done, some styling and API requests missing. Is there an issue to add redux workflow? I think it would be useful to manage store overall, there are few things missing, but let me know, what needs to be fixed here ProgramEquity/product#40 
Screen Shot 2022-03-10 at 1 27 27 PM


mahrsbar:
Vuex store (which is the redux equivalent for Vue) is on the horizon for sure in the upcoming months but I haven't created an issue yet. We are prioritizing the features at the moment. 

Edits look great! You can usually apply [vuetify](https://vuetifyjs.com/en/styles/flex/#enabling-flexbox) classes to centre and align columns instead of adding custom css under `<style>` but its not a big deal.

I think at this moment we're just going to hardcore the cards, so if you can export the images from the [figma](https://www.figma.com/file/46c9cmuTiCpFA4DHB8OK0H/Amplify-User-Interface-%2B-Design-Guide?node-id=2282%3A2531) file and add them, that would be awesome!

Thanks!

joester030376 (commented):
I checked 4 files.
the code content is good.
but can we make the code structure a bit differently.
the page with style sheet inside one file is not good.
They should be divided into the different files individually.

manishapriya94:
@aneelac22 let me know if you have the bandwidth this week to review?

Issue #29, opened 2022-01-26
Batch: Provide photos (and more info) for representatives


JamesMGreene:
We are currently using the Google Civic API to look up relevant state and federal representative information based on the user's address. This works very well! 🎉 

However, the Civic API is also missing quite a bit of information about those representatives, often including their photos and sometimes even information contact information. 😭 

One strategy that we could employ to workaround this shortcoming is to stitch the information provided by the Google Civic API with information provided by other public APIs, such as the Open States database: https://docs.openstates.org/api-v3/ 🔀 

I'm unsure of its completeness (it seems pretty thorough at first glance) but it's definitely _better_ than relying on just the incomplete information from the Google Civic API. 😓

manishapriya94:
@JamesMGreene We should have 5250 credits now and they do use up **one credit for each of these calls**: 
<img width="719" alt="Screen Shot 2022-03-10 at 9 49 51 AM" src="https://user-images.githubusercontent.com/9143339/157725459-cbd6032d-6e39-46cb-a501-0e83148efbe1.png">

The secret is CICERO_KEY and the base url to call is: https://cicero.azavea.com/v3.1

gdomingu:
It sounds like from discussion today we want to move forward with switching over to Cicero api in order to get images.

ahn-nath:
@manishapriya94 

Why isn't this issue closed?

paramsiddharth:
> @manishapriya94
> 
> Why isn't this issue closed?

@ahn-nath Hmmm... This is a research thread, so we can probably close it. Still, I think part of this issue might correspond to #399, in which case it seems more appropriate to keep it open.

PR #127, opened 2022-03-15
[Experiment] Use Open States API instead of the Google Civic API

JamesMGreene:
[Experiment] Use Open States API instead of the Google Civic API

Closes #121

Towards https://github.com/ProgramEquity/amplify/issues/29

:warning: The results for the Open States API definitely seem a lot smaller (fewer reps returned) and more inconsistent on some parts (e.g. office addresses). It is also missing U.S. Senators. 😭  It _does_, however, include a photo URL. 😅


JamesMGreene:
I checked with the Open States folks at Civic Eagle. They suggested we could use this alternative backend call to get the U.S. Senators factored in, e.g.
```
https://openstates.org/find_your_legislator/?lat=44.65&lon=-93.14&address=55024&state=json=json
```

_buuuuut_ it's outside of the official API so no idea on if we will hit rate limiting, and the response schema is very tightly scope such that it's missing details we want like office addresses. 👎🏻 

I'm going to make the call and say that the Open States API, as it is today, does not meet our current needs. 🙅🏻‍♂️

manishapriya94:
@Param Siddharth here are election calendars..wondering if we should almost be using ballotpedia's open API to create flags for refreshing cicero's cache? https://ballotpedia.org/Elections_calendar

cc for visibility: @James Greene this is regarding the block we ran into with the pictures for reps and how we should cache what cicero API returns. @Param Siddharth was speaking to how we can create a reps data structure similar to campaigns API. Ballotpedia would help us understand when to refresh the cache and by jursidiction

paramsiddharth:
@manishapriya94 You'd want to look at this.

Source: https://ballotpedia.org/API-documentation/faq

image

image

JamesMGreene:
Cicero data includes a field that indicates when the official's term is set to expire, which is probably a pretty strong cue to utilize in most cases. Either way, I think the Open States API (not Cicero) is out of the running as far as I can tell. 🙅🏻

paramsiddharth:
> Cicero data includes a field that indicates when the official's term is set to expire, which is probably a pretty strong cue to utilize in most cases. Either way, I think the Open States API (not Cicero) is out of the running as far as I can tell. 🙅🏻

@JamesMGreene Wow, James! Do you think we might be able to use the same while implementing the cache refreshing strategy?

JamesMGreene:
> Do you think we might be able to use the same while implementing the cache refreshing strategy?

It feels like it could be at least a good default to drive refreshes with. 👍🏻 

The field I was referring to is called `valid_to`.

From https://cicero.azavea.com/docs/#official-object:

> `valid_to`: Date the official's term in office expires. This date would be extended if the official wins re-election.

PR #144, opened 2022-03-23
Create database models using Objection ORM

### LINKED ISSUE #32: Epic: ORM for database modeling and migrations
My experiences with Sequelize have all become cautionary tales, so I guided us away from using it at this point and instead favored just using Knex, which is a more basic "SQL query builder".

However, for maintainability's sake, it is probably worthwhile to consider an ORM for the long-term, especially with some contributors not being too familiar with databases.

~~I would suggest we investigate [Prisma](https://www.prisma.io/) as a potential ORM in particular.~~ 🔺

~~I think it has a lot of nice features that make it a more solid choice than Sequelize. It should also be able to completely replace our Knex usage, as far as I can tell at this point.~~

### Tasks

- [x] Install Objection in the codebase: `npm install --save objection`
- [x] Configure Objection in the codebase
  - This requires a one-time configuration something like `require('objection').Model.knex(knexClient)`
- [x] https://github.com/ProgramEquity/amplify/issues/141
- [x] https://github.com/ProgramEquity/amplify/issues/142
- [ ] **BONUS:** https://github.com/ProgramEquity/amplify/issues/143 (:warning: currently unused)
- [ ] **BONUS:** Create a Model for the `constituents` table (:warning: currently unused)
- [ ] **BONUS:** Create a Model for the `sent_letters` table (:warning: currently unused)
- [ ] **BONUS:** Create a Model for the `variants` table (:warning: currently unused)

### Status

- There is a branch created called [`objection-orm`](https://github.com/ProgramEquity/amplify/tree/objection-orm) that installs the `objection` module and sets up the project for creating models.
- Feel free to create branches off that branch and send PRs to it as the base branch
- Models should be created in the `server/db/models/` directory (only exists in the `objection-orm` branch)
- Model filename should be in a singular (not plural) form in all lowercase letters with hyphens as needed, e.g. `letter-version.js`
- Model class name should in a singular (not plural) form in PascalCase letters, e.g. `LetterVersion`

Example, at filename `server/db/models/letter-version.js`:

```js
const Model = require('./_base')

class LetterVersion extends Model {
  static get tableName() {
    return 'letter_versions'
  }

  // ...additional methods, relationships, etc.
}

module.exports = LetterVersion
```

JamesMGreene:
Closes #32

### DIFF
--- a/server/routes/api/letter_versions.js
+++ b/server/routes/api/letter_versions.js
@@ -1,15 +1,14 @@
 const express = require('express')
 const router = express.Router()
-const Campaign = require('../../db/models/campaign')
-//const LetterVersion = require('../../db/models/letter-version')
+const LetterVersion = require('../../db/models/letter-version')
 
 router.get('/:campaignId', async (req, res) => {
   const campaignId = req.params.campaignId
   try {
-    const letterVersions = await Campaign.relatedQuery('LetterVersions').for(
+    const letterVersions = await LetterVersion.query().where(
+      'campaign_id',
       campaignId
     )
-    //const letterVersions = await LetterVersion.query().where('campaign_id', campaignId)
     res.send(letterVersions)
   } catch (error) {
     console.log(error)
--- a/server/routes/api/letter_versions.js
+++ b/server/routes/api/letter_versions.js
@@ -6,9 +6,11 @@ const Campaign = require('../../db/models/campaign')
 router.get('/:campaignId', async (req, res) => {
   const campaignId = req.params.campaignId
   try {
-    const result = await Campaign.relatedQuery('LetterVersions').for(campaignId)
-    //const result = await LetterVersion.query().where('campaign_id', campaignId)
-    res.send(result)
+    const letterVersions = await Campaign.relatedQuery('LetterVersions').for(
+      campaignId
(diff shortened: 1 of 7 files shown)


server/routes/api/letter_versions.js
```diff
@@ -1,12 +1,13 @@
...
-const db = createClient()
+const Campaign = require('../../db/models/campaign')
+//const LetterVersion = require('../../db/models/letter-version')
 
 router.get('/:campaignId', async (req, res) => {
   const campaignId = req.params.campaignId
   try {
-    const result = await db('letter_versions').where('campaign_id', campaignId)
+    const result = await Campaign.relatedQuery('LetterVersions').for(campaignId)
+    //const result = await LetterVersion.query().where('campaign_id', campaignId)
```
JamesMGreene, line 10: These both generate the same underlying SQL query. Which feels more readable? 🤷🏻‍♂️

manishapriya94:
@JamesMGreene ok to merge this for #49?

JamesMGreene:
@manishapriya94 Yup! Allowing time for potential reviewers but looks like it's just time to move ahead. 👍🏻

PR #154, opened 2022-03-23
Modify postal code submit flow

joshjohanning:
Problem:
- Clicking "enter" after you typed in your zip code just refreshed the page

Solution: 
- Tried to get enter to not refresh the page but that did not work. Instead, disabled enter from doing anything.
- If valid zip is entered, the results will appear on the screen automatically (no enter or submit button push)
- Remove submit button on postal code screen

https://user-images.githubusercontent.com/19912012/159782076-38d47023-5f44-4cf7-8aa2-1c029b2ed790.mov

### DIFF
--- a/src/components/SearchReps.vue
+++ b/src/components/SearchReps.vue
@@ -8,9 +8,14 @@
 
             <v-form ref="form">
               <v-text-field
+                :to="{
+                  name: 'Reps',
+                  params: { postalCode: postalCode }
+                }"
+                v-on:keyup="CreateRepList()"
+                v-on:keydown.enter.prevent="CreateRepList()"
                 label="Postal Code"
                 required
-                v-on:keyup="CheckInputContent"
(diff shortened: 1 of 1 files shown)


mahrsbar (changes requested):
This is a great catch and definitely needed, but I think it might be affecting the URL that is built once the user enters a postal code. 

The goal is to have the url as "postalcode/xxxxx/member/xxxxxx" as the user progresses so once they log in they are redirected back to where they were. This might have been broken though. 

I would also include an error message if the postal code is incorrect ("Invalid postal code") which I think was added in the back end but might not be working correctly.

manishapriya94:
@teakopp did we configure the error message here?

joshjohanning:
> @teakopp did we configure the error message here?

@manishapriya94 I don't believe it showed anything as of last Friday for an error message on an invalid zip.

Issue #141, opened 2022-03-23
Create a Model for the `campaigns` table


JamesMGreene:
Create an Objection Model for the `campaigns` table
- Update its usage in `server/routes/api/campaigns.js`
- Consider adding any practical helper/accessor behavioral methods that seem relevant to our usage

Towards #32

JamesMGreene:
Closed by https://github.com/ProgramEquity/amplify/pull/144

PR #156, opened 2022-03-23
Added campaign orm model. Modified API

mickeygousset:
This PR adds an ORM for the `campaigns' table.  Issue #141 

I created a new object model for `campaigns`, then modified the `campaigns` api to use this object model.  I did not see where the campaign table had any relations to other tables, so I did not create any relations in the ORM.

### DIFF
--- a/server/db/models/campaign.js
+++ b/server/db/models/campaign.js
@@ -1,0 +1,29 @@
+const Model = require('./_base')
+
+class Campaign extends Model {
+  static get tableName() {
+    return 'campaigns'
+  }
+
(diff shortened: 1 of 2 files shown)


manishapriya94:
@JamesMGreene or @waldnzwrld, would you have the bandwidth to review?

manishapriya94:
@nawazkhan would you be able to review?

JamesMGreene:
@mickeygousset Apologies that it took so long to get this merged, but we do really appreciate your contribution! 💝

Issue #19, opened 2022-02-08
Hack Pod: Enhance Dev Experience via Codespaces + Secure Workflows 


manishapriya94:
⬇️ **Table of Contents:** 
- [User Story: Who does this Affect](#user-story)
- [Current Contribution Flow and Actions Checklist](#workflow)
- [Codespaces Specs and Tasks](#codespaces)
- 💻[Project view](https://github.com/orgs/ProgramEquity/projects/8)

## User Story <a id = "user-story"> </a>
<img width="510" alt="Screen Shot 2022-04-14 at 2 49 53 PM" src="https://user-images.githubusercontent.com/9143339/163482322-539fbcfc-6523-4b62-9b4b-decadb3f52b6.png">

=- Our goal is to enable students towards enterprise architecture while making their onboarding painless 

## [Current Contribution Flow and Actions](https://github.com/ProgramEquity/amplify/discussions/61) <a id = "workflow"> </a>
Today we use Actions to perform administrative automations, CI/CD, and use it for CodeQL (in pink)

![image](https://user-images.githubusercontent.com/9143339/195364390-e9dc0efd-0ea4-4d5e-a2e5-9c0666822349.png)

- [Current workflows](https://github.com/ProgramEquity/amplify/actions)
  - checks formatting
  - Lint code
  - Lint workflows
  - Scorecards
  - CodeQL
  - Integration tests

### Considerations 
- set up default container configuration
- Adding additional features to your `devcontainer.json` file
- How can we make it easier for outside collaborators that are new to open source provide their own API keys as secrets in order for the environment variables in codespaces to work? 
  - Currently, this creates a barrier for new students who also have to learn concepts of APIs keys, and secrets for each API we use by going to each  (4: Lob, Auth0, Google Civic, Stripe). What strategies do we want to have for people to route Auth0 (Even if the create their own key, how do we allow for that as a variable?
  - In particular, Auth0 callback is hard to configure for codespaces domains without each contributor having to create their own Auth0 account
- Are there container specs that are needed?
- Should we be adding any external lints for Docker?
- Bonus: what gaps can the readme fill?

**Resources:** 
- [Try out our reference](https://github.com/ProgramEquity/amplify-back-end/pull/132)
- [Contribution Guide](https://github.com/ProgramEquity/amplify/blob/main/.github/CONTRIBUTING.md)
- [GitHub Docs on Codespaces](https://docs.github.com/en/codespaces)
[...]

PR #157, opened 2022-03-24
Configure Codespaces and Development Containers

kenmuse:
*Problem:*
- The Codespaces environment was not fully configured to enable developers use the environment as-configured. This is a continuation of the feature requested in #19.
- The environment did not work when used for local development. This prevented teams without access to Codespaces from using a local VS Code development container.

*Solution:*
- Configured `devcontainer.json` to ensure the setup scripts have execute permission in the container before running. Also modified the file to automatically copy `.env.example` for the users to ensure that the npm commands succeed. Developers can immediately start developing by typing `npm run dev`
- Added a `.dockerignore` to ensure that `node_modules` and `.env` are not included in the dev containers
- Configured 'docker-compose.yml' to create an anonymous volume for the `node_modules`.
- Configured `Dockerfile` with a defined `node_modules` folder in the workspace with the owner specified as `node:node`. This avoids a permissions error with the anonymous volume.
- Configured `package.json` (and `package-lock.json`) to update the Husky install command to use `npx` to ensure it executes properly in a local container. This sets up the hooks.


manishapriya94:
Thanks @kenmuse, @JamesMGreene would you like to review?

kenmuse:
Just validated that this container will run correctly on an M1-based Mac in addition to x86 PCs and Codespaces.

kenmuse:
The `devcontainer.json` schema has changed. Some minor changes are required to update that to match the latest proposal.

kenmuse:
I'm checking to see whether there is still an issue with Windows dev containers that may require the `chmod` to be effective. In the past, if you started the dev container from the Windows file system in it's directories, the bit could get lost when mounting into the container. However, if you cloned the repo inside $HOME in WSL, this would not happen (since Unix preserves the bits, allowing them to mount in the container).

paramsiddharth:
@kenmuse Hmmm… Well, if you start the devcontainer using Docker, it doesn't use the Windows filesystem. Instead, it uses a WSL2 filesystem within distros registered as `docker-desktop` and `docker-desktop-data`. Even outside Docker, the recommended place to develop in WSL2 is inside the Linux filesystem and not via the interface to Windows' FS. Also, the devcontainer runs via a Linux image, so I don't think that the executable bit will be lost. What are your thoughts? :)

paramsiddharth:
> Since the `devcontainer.json` file is going to initialize a prepare container for the development based on the base image specified inside the Dockerfile/specification, and Docker on Windows by default utilizes either a WSL2 or Hyper-V back-end, I don't think we need to keep the commands platform-agnostic (e.g. the `chmod +x` parts) because we plan to spin up Linux containers anyway.

@kenmuse See this, too.

PR #188, opened 2022-04-29
Support for custom Campaign logos

JamesMGreene:
Adds support for custom Campaign logos. Can't say I'm super proud of it but it's functional enough for the short-term... hosting Campaign logo images locally may not scale well long-term. 🤷🏻‍♂️ 

Example visual with some local data:

image

Example values in the database:
- `NULL` loads the default image
- paths starting with `/` assume they're local images in our `public/` directory
- anything else is assumed to be an external site (or other valid `img src` attribute value) and just allowed to load 🤷🏻‍♂️ 

image


JamesMGreene:
Confirmed with @manishapriya94. 🚀

JamesMGreene:
In production:

image

PR #216, opened 2022-05-24
Stripe implementation

joester030376:


server/__tests__/integration/lob.test.js
```diff
@@ -353,9 +353,13 @@ describe('POST /api/lob/createLetter', () => {
     }
 
     const template_id = 'tmpl_1057bb6f50f81fb'
+
+    // A test payment intent that should return status of 'succeeded'
+    const paymentIntentId = 'pi_3L7VXGFqipIA40A31qbflVvO'
```
paramsiddharth, line 358: Are these values public keys that need to be fixed in staging and development? Is there any specific reason why we must hardcode them instead of using environment variables?

manishapriya94: @paramsiddharth I think @DietBepis1 was adding dummy variables to study behavior

PR #217, opened 2022-05-25
Feature/complete page

serunkuma:
with this pull request. we should be closing PR: https://github.com/ProgramEquity/amplify/pull/126

it adds onto it and makes it work for the complete page.


serunkuma:
127 0 0 1_8080_complete

this is how it looks like for the complete page

PR #241, opened 2022-06-24
Re-enable CORS support but ONLY for developers

JamesMGreene:
Allow CORS support but only for development scenarios (e.g. to enable Postman testing)

Mostly reverts commit 662d763139e37820cc317b357f9a27ffedbec206

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "axios": "^0.27.2",
         "cli-plugin-eslint": "^0.1.0",
         "connect-history-api-fallback": "^1.6.0",
+        "cors": "^2.8.5",
         "dotenv": "^16.0.0",
         "express": "^4.18.1",
         "express-jwt": "^6.1.1",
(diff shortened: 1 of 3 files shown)


manishapriya94 (approved):
Might need to change our tests action file for the future

paramsiddharth (approved):
The PR looks good to me. However,

@JamesMGreene Do we really need this though? I think Postman should still work without adding CORS support. CORS is only needed for front-end browser requests, isn't it? Why else would we want to enable wildcard access under CORS in development environments?

References:
1. [_Postman does not implement the CORS restrictions, which is why you don’t see the same error when making the same call from Postman._](https://community.postman.com/t/no-access-control-allow-origin-header-is-present-on-the-requested-resource/22199/5)
2. [_Postman doesn't care about SOP, it's a dev tool not a browser_](https://stackoverflow.com/questions/36250615/cors-with-postman)

JamesMGreene:
@paramsiddharth Good point. (For history: Postman used to be a browser extension rather than a standalone app like it is today.)

I'd be curious how other browser-based HTTP/REST clients like Insomnia react without CORS for accessing a development backend. 🤔 

Now that you bring it up, I think we've primarily run into this when devs were using Postman but targeting our production deployment (which _should_ definitely get blocked) rather than their own development server. 💭

JamesMGreene:
@paramsiddharth Thanks for doing the research to confirm! ✨

PR #242, opened 2022-06-28
Bug: Backend Stripe Verification

### LINKED ISSUE #213: Bug:Create_letter needs a backend stripe verification
We have a buildsafe bug. Currently, Stripe checkout flow is implemented as frontend only (payment window opens once pay and send letter is clicked by user)

## Expected Behavior
Screen Shot 2022-05-04 at 1 35 33 PM
Screen Shot 2022-05-04 at 1 36 45 PM
We want the letter object to only be created once payment token being successful has returned 

## Current Behavior

When the user hits send button, a letter object is being created regardless of payment. This might be because checkout flow is on frontend but not being validated in backend. In frontend, the successful action won't load until payment is processed but since there isn't a check for backend, the letter object has already been posted. We need the backend to validate the stripe before the object is posted.

## Possible Implementation
In create_letter function in lob.js
In the try block, we need the stripe client to be verified

Reference: https://stripe.com/docs/api/charges/retrieve - to verify a Stripe charge on the backend

paramsiddharth:
Resolves #213. Few points to be noted:

1. I am unable to make sense of this setting, nor was keeping it letting the success page show the e-mail address and donated amount correctly. It was showing an error in the browser console too, due to which I removed it.

   `src/components/ActionComplete.vue#30`
   ```js
   methods: {
       getSession() {
           axios.defaults.baseURL = '//localhost:6000/';
           //                       ^^^^^^^^^^^^^^^^^^^^
           const session_id = {'session_id': this.$route.query.session_id }
           axios.post( '/api/checkout/create-transaction', session_id )
               .then((response) => {
                   this.data  = response.data
               })
               .catch(function (error) {
                   console.log(error)
               })
         },
   }
   ```

2. I can't find anywhere in the front-end where `/createLetter` is being `POST`ed (except for the test case in `server/__tests__/integration/lob.test.js`). Does that mean the letter-sending end-point is not being called yet? 

   I modified it to return the failure status code (as received from Stripe) so that it can be dealt with in the front-end, when we implement the part where it is being called.

3. A failed payment may never lead us to the success page or the back-end, because the Stripe window won't let me proceed _until_ the payment finally succeeds. This conclusion comes by testing the [failing test cards](https://stripe.com/docs/testing#declined-payments) as discussed in last week's pair hours.

I think this PR needs review and further discussion.

### DIFF
--- a/server/routes/api/lob.js
+++ b/server/routes/api/lob.js
@@ -158,9 +158,13 @@ router.post('/createLetter', async (req, res) => {
     )
 
     if (paymentVerification.status !== 'succeeded') {
+      const { status } = paymentVerification
       return res
         .status(500)
-        .json({ msg: 'Payment is still pending or has failed.' })
+        .json({
+          msg: 'Payment is still pending or has failed.',
+          status
+        })
         .end()
     }
 
--- a/src/components/ActionComplete.vue
+++ b/src/components/ActionComplete.vue
@@ -27,7 +27,6 @@ export default {
     },
     methods: {
         getSession() {
-            axios.defaults.baseURL = '//localhost:6000/';
             const session_id = {'session_id': this.$route.query.session_id }
             axios.post( '/api/checkout/create-transaction', session_id )
                 .then((response) => {


thyeggman:
1. @paramsiddharth What was the error that you were getting when the `axios.defaults.baseURL` setting was present?  I will try to see if I can repro later tonight
2. You're right, it doesn't look like the `createLetter` endpoint is being used at all on the front end, likely for the reason that you mention.  Maybe @manishapriya94 has some insight into why it hasn't been implemented yet, and what the plans are to use it?
3. Kind of relies on 2, but it might be worth looking at [webhook postbacks](https://stripe.com/docs/webhooks#webhook-endpoint-one) in the stripe docs

paramsiddharth:
> 1. @paramsiddharth What was the error that you were getting when the `axios.defaults.baseURL` setting was present?  I will try to see if I can repro later tonight
> 2. You're right, it doesn't look like the `createLetter` endpoint is being used at all on the front end, likely for the reason that you mention.  Maybe @manishapriya94 has some insight into why it hasn't been implemented yet, and what the plans are to implement it in a safe way that doesn't incur charges in testing?
> 3. Kind of relies on 2, but it might be worth looking at [webhook postbacks](https://stripe.com/docs/webhooks#webhook-endpoint-one) in the stripe docs

1. When the setting was present, no amount was being displayed on the confirmation page and the browser console had an `AxiosError`, which is expected since there's nothing on port 6000 in localhost.
2 and 3. If the `/createLetter` endpoint is being used as a webhook, I would like to see a demo of the Lob dashboard to see what it looks like and how we are configuring the webhook signal (because it is clearly not being done in the code). Also, failed payments will _NOT_ call the webhook, because the payment page of Stripe doesn't let us continue without the payment succeeding.

paramsiddharth:
Hey @DietBepis1 ! Could you tell us if `/createLetter` is meant for the front-end or if it is registered as a webhook in Stripe (or something else)?

DietBepis1:
Hey @paramsiddharth,  as far as I have been able to tell, /createLetter should be called in the ActionComplete component after a successful redirection from Stripe checkout. I am working on building that page out currently and will have a PR for it in the couple days. I have not seen any other instances of it called in the codebase (besides the integration test, as you mentioned).

Related to this, I realized the other day that the way I intended the back-end verification to work with the Stripe paymentIntentId will need the Stripe sessionId from the redirect as a parameter in /createLetter. That way you can retrieve paymentIntent from the session object. I have fixed it in the branch that I am using for the completion page, but I can fix in this other branch if that is better.

I also get an error from that axios.defaults.baseUrl and just deleted it from the completion page, since it seems like the API calls should actually be using the same base url as the application (maybe this is wrong for production, please let me know).

I have a call set up with @manishapriya94 tomorrow afternoon at 5pm central and these were on my agenda to talk about. I am happy to have anyone else join too. I can email you the link or we can have a Slack meeting. Just let me know!

paramsiddharth:
> Hey @paramsiddharth, as far as I have been able to tell, /createLetter should be called in the ActionComplete component after a successful redirection from Stripe checkout. I am working on building that page out currently and will have a PR for it in the couple days. I have not seen any other instances of it called in the codebase (besides the integration test, as you mentioned).
> 
> Related to this, I realized the other day that the way I intended the back-end verification to work with the Stripe paymentIntentId will need the Stripe sessionId from the redirect as a parameter in /createLetter. That way you can retrieve paymentIntent from the session object. I have fixed it in the branch that I am using for the completion page, but I can fix in this other branch if that is better.
> 
> I also get an error from that axios.defaults.baseUrl and just deleted it from the completion page, since it seems like the API calls should actually be using the same base url as the application (maybe this is wrong for production, please let me know).
> 
> I have a call set up with @manishapriya94 tomorrow afternoon at 5pm central and these were on my agenda to talk about. I am happy to have anyone else join too. I can email you the link or we can have a Slack meeting. Just let me know!

Hey @DietBepis1 ! Noted, so it is not being called as a webhook.

As far as the changes and fixes are concerned, it would be better we have them merged with your upcoming PR and I can build on top of them.

The production application is running on Heroku, and if I am understanding it right, it still uses the same back-end front-end pattern as we are using once the application is built i.e. Serving the back-end on the given `$PORT` and statically serving the compiled Vue app. But yes, please update me if we find a real use of that `axios.defaults.baseURL` line.

5 PM CT might be a little too early for me, but I'll definitely try to attend! Here's my e-mail: me@paramsid.com. Please send me the meeting link, and the calendar invite too if possible. If I miss we can try meeting separately via Slack. Thank you!

PR #244, opened 2022-06-29
Updates script/bootstrap so it works on codespace launch

gdomingu:
Noticed that the `script/bootstrap` would fail on the `brew` steps because brew wasn't being added to the path when creating it from a codespace. This should fix that. 

Also updates package.json because was complaining about not being able to resolve less-loader@5.0.0 for some other packages.


paramsiddharth (commented):
Hey @gdomingu ! Just a question here — Did you regenerate the lockfile here or did you modify it directly?

gdomingu:
> Hey @gdomingu ! Just a question here — Did you regenerate the lockfile here or did you modify it directly?

@paramsiddharth I ran `npm install --legacy-peer-deps` and it regenerated it for me.

paramsiddharth (dismissed):
Alrighty. Looks good to me!

JamesMGreene:
@gdomingu Ideally, running `script/bootstrap` shouldn't be necessary for Codespaces if we setup the container correctly. 😅  Did you find that it was needed for you?

JamesMGreene (approved):

PR #288, opened 2022-08-09
Add workflow to verify clean setup for Codespaces

JamesMGreene:
Adding a check to verify a hopefully clean setup for Codespaces. Doing so after running into issues with `npm install` hitting issues for peer dependency conflicts both in Codespaces and when deploying to Heroku, but NOT hitting them while running our other CI builds.

Primary difference: utilizing npm caching and `npm ci` vs. no cache and `npm install`.

The Codespaces setup script will also verify that the local developer (Codespaces) experience for database bootstrapping is not failing either. 👍🏻

### DIFF
--- a/.github/workflows/verify-clean-setup.yml
+++ b/.github/workflows/verify-clean-setup.yml
@@ -1,0 +1,39 @@
+name: Verify clean setup for Codespaces
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


JamesMGreene:
Nope, not going to pan out. There's no database present to work with. 😓

PR #346, opened 2022-10-12
Add the cats gif on pr push workflow

Lehcar:
Co-authored-by: Manisha Priyadarshini <manishapriya94@github.com>

## Description

This adds the [action-cats workflow](https://github.com/ruairidhwm/action-cats) to the amplify repo.

### DIFF
--- a/.github/workflows/cats.yml
+++ b/.github/workflows/cats.yml
@@ -1,0 +1,22 @@
+name: Add a cat gif on pr push
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+
+permissions:
+  contents: read
+
+jobs:
+  cats:
+    name: Cat
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Action Cats
+        uses: ruairidhwm/action-cats@1.0.2
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/cats.yml
```diff
@@ -0,0 +1,17 @@
...
+      - reopened
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      run: echo "Hello World"
```
paramsiddharth, line 17:
```suggestion
      - name: Print "Hello World"
        run: echo "Hello World"
```

paramsiddharth (changes requested):
Hi @Lehcar ! I think this might be an issue with the workflow preventing it from running.

Lehcar:
I'm going to close and reopen this PR to try and trigger the workflow (instead of adding the workflow_dispatch)

Lehcar:
Ok, this might be an issue with permissions, going to debug a bit

paramsiddharth:
@Lehcar I think reopening the PR might not trigger the workflow yet, just because the workflow isn't merged into `main`? So I think we can continue relying on `workflow_dispatch` and try to debug the workflow till then. :)

Lehcar:
> 

the actions UI isn't recognizing the action even with the workflow_dispatch trigger set. 
I'm not sure why, it should be appearing there 😕

Lehcar:
going to try and trigger this action by closing and opening? :?

Lehcar:
So we think the problem was that the Check formatting check was failing, causing the action to exit before it could pickup this workflow file. 

I pushed up a commit with the workflow_dispatch trigger on and with that check passing and the action ran (it did fail because it couldn't find the PR): https://github.com/ProgramEquity/amplify/actions/runs/3237828415

This _should_ work, but it can't really be tested until it's merged

paramsiddharth:
> So we think the problem was that the Check formatting check was failing, causing the action to exit before it could pickup this workflow file.
> 
> I pushed up a commit with the workflow_dispatch trigger on and with that check passing and the action ran (it did fail because it couldn't find the PR): https://github.com/ProgramEquity/amplify/actions/runs/3237828415
> 
> This _should_ work, but it can't really be tested until it's merged

Hmmm... @manishapriya94 What do you think? Should we go ahead and merge it to test it further?

PR #358, opened 2022-10-13
Added Table of Contents

LoopFruits:
Easy for users to navigate to different sections.

Documentation only PR, no builds affected


README.md
```diff
@@ -1,5 +1,16 @@
...
+==================
+
+  * [Getting Started](#getting-started)
+  * [Project Setup](#project-setup)
+  * [Customize configuration](#customize-configuration)
+  * [App Walkthrough](#app-walkthrough)
+  
+  
+  
+
```
paramsiddharth, line 13:
```suggestion
```

README.md
```diff
@@ -1,5 +1,16 @@
 # amplify
 
+Table of Contents
+==================
```
paramsiddharth, line 4:
```suggestion
## Table of Contents
```

README.md
paramsiddharth, line 13:
```suggestion
```

paramsiddharth (changes requested):
There are some minor changes needed for consistency. The rest looks good!

paramsiddharth (approved):

JimmyJammer:
FWIW, GitHub automatically generates a Table of Contents for these if you just click on the upper-left menu button near the filename in the viewing panel:

Screen Shot 2022-10-20 at 10 47 41 AM

The beauty of GitHub's automatic solution is that it will never become out-of-date. ✨ 

With that in mind, do we still want to add this? It is undoubtedly more visible, so it's just a question of if that visibility outweighs the long-term maintenance burden. 🤷‍♂️ 🤔

paramsiddharth:
> @JimmyJammer FWIW, GitHub automatically generates a Table of Contents for these if you just click on the upper-left menu button near the filename in the viewing panel:
> 
> Screen Shot 2022-10-20 at 10 47 41 AM
> 
> The beauty of GitHub's automatic solution is that it will never become out-of-date. ✨
> 
> With that in mind, do we still want to add this? It is undoubtedly more visible, so it's just a question of if that visibility outweighs the long-term maintenance burden. 🤷‍♂️ 🤔

Good point! As amazing as this feature is, I think we'd still go with considering this PR for this particular repository. "More visibility" isn't exactly going to be the criterion, but I think the design team intended on having this added to the README.

Nevertheless, thank you for pointing out the fact that there's a button that automatically generates the ToC for us! It will now be helpful for me every time I navigate through GitHub repositories. 🤗

JamesMGreene:
Forgot I was logged into my test account there. 😊

paramsiddharth:
> Forgot I was logged into my test account there. 😊

@JamesMGreene Haha, big reveal, James! 😁 Your test account just got a new follower~! Had I known it was you, I'll probably let you make the final decision instead of what was proposed in the bugbash. Sorry about that!

JamesMGreene:
I'm fine with the change! 👍🏻

Issue #332, opened 2022-10-12
Fix the display of the filter in the campaign page form


paramsiddharth:
# Fix the display of the filter in the campaign page form
The filters are often overlapped by the submit button.

![image](https://user-images.githubusercontent.com/30315706/195389184-c600c7cc-c3f1-45b6-a50b-4556576dc178.png)

## Solution
- Update the styling in the relevant components to avoid the elements overlapping.
- Test out the update on different Windows sizes and device modes (using the browser inspection tool).

paramsiddharth:
@ahn-nath 
![image](https://user-images.githubusercontent.com/30315706/196694025-531b93a4-d1ea-4faf-808f-29ed0d6b86c3.png)
Tip: Instead of writing "Issue: #332" like you did, you can write "Resolves #332" or "Closes #332" to automatically link the PR to the issue so that the issue gets closed once the PR is merged. More on that here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

ahn-nath:
> @ahn-nath ![image](https://user-images.githubusercontent.com/30315706/196694025-531b93a4-d1ea-4faf-808f-29ed0d6b86c3.png) Tip: Instead of writing "Issue: #332" like you did, you can write "Resolves #332" or "Closes #332" to automatically link the PR to the issue so that the issue gets closed once the PR is merged. More on that here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Awesome!

PR #365, opened 2022-10-17
Fix the display of the filter in the campaign page form

ahn-nath:
Issue: #332 

**Solution:** 
The button element is not the child of any wrapper component with proper margin and padding. The same is true for the buttons over it. The behavior, therefore, was hard to predict, and it was on top of items with similar characteristics. I was going to add a default margin to the button, however, after checking the code, I realized wrappers would be enough and proportionate more control over the element. 

I have also added columns to the elements wrapped by a row without them. The logic for it was based on the grid system of Vuetify: ([Grid system — Vuetify](https://vuetifyjs.com/en/components/grids/#v-container)), where is stated that v-row is a wrapper component for v-col that utilizes flex properties to control the layout and flow of its inner columns. It is more consistent and logical to use the row with a column, as the class contains several properties that are specific to columns. 

A demo that shows how the change looks in different window resolutions/sizes: https://www.loom.com/share/b90eb61e1df84ba8849347093a72fd57  

Looking forward to your suggestions and comments. 

**References:** 
Vuetify | A Material Design Framework for Vue.js. (n.d.). Retrieved October 17, 2022, from https://vuetifyjs.com/en/components/grids/


src/components/SearchReps.vue
```diff
@@ -16,71 +16,84 @@
...
+                  }"
+                  v-on:click="CreateRepList()"
+                  clickclass="mr-4"
+                  >Submit
+                </v-btn>
+              </v-col>
+            </v-row>
           </v-card-text>
         </v-card>
+
```
paramsiddharth, line 96: Here, if we could put all the buttons (with proper margins amongst themselves) within the same column, we wouldn't see a lot of gap between them as we see here:

image

As the window size keeps increasing, the gap between the filter buttons keeps increasing too. It would be better if the gap remains consistent and if all the buttons are stored within a single column contained in the row component.

paramsiddharth (changes requested):
@ahn-nath Excellent work, Nath! For someone just a few weeks into the fellowship and one who's new to Vue.js, you're performing great. I really appreciate the efforts and quality of description on the PRs. It makes reviewing them a tonne easier for the reviewers. Absolutely loved it!

I've requested one change. Please look at it and go ahead with it. Once that's done I think the PR will be ready for approval and merge. :)

ahn-nath (commented):
I agree with the changes and understand the point of using one column for all buttons.

ahn-nath (commented):
Good @paramsiddharth. I think we can merge the changes now.

paramsiddharth (changes requested):
Hi, Nath. I would suggest you revert the PascalCase update because it's just a warning and we can ignore it for now. I want us to continue using the current convention for the Font Awesome icon component while still being able to ignore the error somehow. There are some known ways to ignore lint errors in certain instances. I'll try to find what we can do to avoid it here.

The rest seems good to me. The screenshots look amazing. Good job!

src/main.js
```diff
@@ -23,7 +23,9 @@ import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 
 /* import specific icons */
 library.add(faUserSecret, faInstagram, faFacebookF, faTwitter, faYoutube)
+
 /* add font awesome icon component */
+// eslint-disable-next-line vue/component-definition-name-casing
```
paramsiddharth, line 28: @ahn-nath I'd like you to look at this. This allows us to ignore the lint errors for the next line. For exceptional cases, this is what we use.

Pro tip: If you're not able to commit/push your changes due to link errors failing in the push/commit hook, you can always use `git commit --no-verify` or `git push -n` to temporarily skip the hooks.

paramsiddharth (approved):
The PR looks good to me now. Feel free to merge it once the tests pass. Good work, Nath!

ahn-nath:
@paramsiddharth 
Could you please merge? 

I am not authorized to do so. Thanks.

PR #366, opened 2022-10-21
Fixed Campaign Page Responsivity for Mobile

sammychinedu2ky:
Fix for https://github.com/ProgramEquity/amplify/issues/334
image

image
@paramsiddharth

Issue #294, opened 2022-08-10
Explore caching options for calling Paid APIs (Cicero)


thyeggman:
The cicero API is billed on a crediting system.  Each API call costs a certain number of credits.  Some APIs are free to use, but as Amplify scales, the cost of using those APIs can become very high.

We are using [axios](https://axios-http.com/docs/intro) to make requests to these APIs.  Axios has a number of different libraries that support caching:

- https://www.npmjs.com/package/axios-cache-adapter
- https://axios-cache-interceptor.js.org/#/

Investigate these libraries to see what would be the best option to cache API calls to reduce the number of credits we use.

ahn-nath:
Hello, @thyeggman, @paramsiddharth. Just to make sure I understand, does the requirement of this issue involves implementing the solution I researched/found? Or would that be out of the scope of this issue?

ahn-nath:
Related to #29 and #270 [logic for using the Cicero API].

paramsiddharth:
> Hello, @thyeggman, @paramsiddharth. Just to make sure I understand, does the requirement of this issue involves implementing the solution I researched/found? Or would that be out of the scope of this issue?

@ahn-nath Yes, I think this issue requires some research. @thyeggman Jacob may be better able to explain to you what he discovered. It would definitely be in scope of this issue to research more, because we want to come up with a robust plan for the issue first before actually starting to code it. What do you suggest, Nath?

ahn-nath:
> > Hello, @thyeggman, @paramsiddharth. Just to make sure I understand, does the requirement of this issue involves implementing the solution I researched/found? Or would that be out of the scope of this issue?
> 
> @ahn-nath Yes, I think this issue requires some research. @thyeggman Jacob may be better able to explain to you what he discovered. It would definitely be in scope of this issue to research more, because we want to come up with a robust plan for the issue first before actually starting to code it. What do you suggest, Nath?

Yes, Param. Thanks for responding, but my question is: does solving this issue specifically requires implementing the solution of the research? That is, coding it? Certain keywords and statements (such as the last paragraph of the description text) and the “explore” of the title confused me a little in regard to the objective.

paramsiddharth:
> > > Hello, @thyeggman, @paramsiddharth. Just to make sure I understand, does the requirement of this issue involves implementing the solution I researched/found? Or would that be out of the scope of this issue?
> > 
> > 
[...]

PR #367, opened 2022-10-23
Ahn nath/crop representatives photo: [Low Priority] Crop representative photos using Cicero API data

### LINKED ISSUE #333: [Low Priority] Crop representative photos using Cicero API data
# Enable cropping of the representative photos using the data provided by the Cicero API
> ⚠️ This issue is extremely low priority and must only be taken if other issues are already taken.
The Cicero API returns profile picture crop data for some representatives. That data can be reformatted and returned by the back-end to the front-end to do extra CSS formatting and present the profile pictures better.

Source: https://cicero.azavea.com/docs/#official-object

image

## Checklist

🖼️: Front-end
⚙️: Back-end
🏗️: Architecture

- [ ] ⚙️ Examine the photo_cropping property.
- [ ] ⚙️ Plan out any modifications that need to be carried out before sending it to the front-end.
- [ ] 🖼️ Apply inline styling to the photos being displayed in individual representative cards to crop them.

ahn-nath:
Closes #333 

In order to take advantage of the photo_cropping property that lets us improve the appearance of member/profile's pics via the API responses, we have:
- Examined the photo_cropping property.
- Planned out any modifications that need to be carried out before sending it to the front-end.
- Applied inline styling to the photos being displayed in individual representative cards to crop them.
- Documented and cleaned the code. 

**Approach:**
The official documentation describes the photo_cropping property as such:
> For photos that are zoomed out or off center, an object of the form {x, y, width, height, origWidth, origHeight} specifying the location (from top left) and dimensions of a bounding box that defines a headshot within the official photo.

It identifies or captures the location of the "ideal" headshot, and they provide the x, and y coordinates of the start of the object/headshot. An example of it: 

` "photo_cropping": {
                                "x": 54,
                                "y": 0,
                                "width": 176,
                                "height": 180,
                                "origWidth": 450,
                                "origHeight": 250
                            }`

We can assume that representatives that do have this property (non-null) have a pic that is either zoomed out or off center. For that reason, instead of actually cropping the pictures and creating a function that makes that update, the code I added simply focuses on the frame or quadrant of the image where the object or headshot starts.

**Consider the following:**
image
 The yellow ball, our object, is at the first quadrant, which means it is at the top-left side of our image of the bigger frame. We locate the object in accordance with its x-location (left or right) and its y-location (top or bottom) and then use the background-position property of the frontend component (v-img) to properly display the pic. 

The logic we use is:
For the x-value or x-location:
```
- If the object/headshot starts at the first part of the left quadrant or side, then we assume the x-value to be left.
- If the object/headshot starts at the first part of the right quadrant or side, then we assume the x-value to be right.
- If none, we assume center as the x-value.
```

For the y-value or y-location:
```
- If the object/headshot starts at the first part of the top quadrant or side, then we assume the y-value to be top.
- If the object/headshot starts at the first part of the bottom quadrant or side, then we assume the y-value to be bottom.
- If none, we assume center as the y-value.
```
Then we simply concatenate both and use them as the CSS value of the background position property. We use the center of the image as a reference for the left and right sides, as well as the top and bottom sides.


ahn-nath:
**Examples:** 
Each example will have the postal code, and the previous (from the production app) and after (the solution of current version of the changes I propose).

**98101**
Andrew J. Lewis 

photo_cropping_wrongt_andrew_l

photo_cropping_correct_andrew_l


**10001**
Mark Levine
photo_cropping_wrong_mark_l

photo_cropping_correct_mark_l

 
**97005**
Tobias Read
photo_cropping_wrong_tobias_r

photo_cropping_correct_tobias_r

ahn-nath:
**Exceptions:** 
If our app does not receive any photo_cropping object (null value), then we cannot modify the pic, even if the change is obvious for us. Consider **Rebecca Rhynhart**, postal code **19147**. We should and can position the image to the right to have a better frame, but the photo_cropping object came as "null".

image
image

server/routes/api/representatives.js
```diff
@@ -115,10 +115,11 @@ router.get('/:zipCode', async (req, res) => {
...
             rep.photo_origin_url ||
             'https://cdn.pixabay.com/photo/2016/08/08/09/17/avatar-1577909_1280.png',
-
-          socialMediaPages: getOfficialSocialMediaPages(rep.identifiers) // call
+          socialMediaPages: getOfficialSocialMediaPages(rep.identifiers),
+          photoCroppingCSS: getPhotoCroppingValues(rep.photo_cropping)
         }
-
+        console.log('rep name:' + repInfo.name)
+        console.log('rep cropping:' + repInfo.photoCroppingCSS)
```
thyeggman, line 122: These should be removed before merging, or guarded by a debug check

server/routes/api/representatives.js
```diff
@@ -129,6 +130,72 @@ router.get('/:zipCode', async (req, res) => {
...
+  }
+
+  // args
+  console.log(photo_cropping_object)
+  let x = photo_cropping_object.x
+  let y = photo_cropping_object.y
+  let oriH = photo_cropping_object.oriHeight
+  let oriW = photo_cropping_object.origWidth
+
+  // 1. calculate threeshold for the x space
```
thyeggman, line 158:
```suggestion
  // 1. calculate threshold for the x space
```

server/routes/api/representatives.js
```diff
@@ -129,6 +130,72 @@ router.get('/:zipCode', async (req, res) => {
...
+  // we check if the coordinate starts on the left side of the image (the first half of the left side)
+  let x_left_threeshold = oriW / 2 / 2
+  // we check if the coordinate starts on the right side of the image (the first half of the right side) and we reduce a margin of 5% to be flexible
+  let x_right_threeshold = oriW / 2 - (5 / 100) * (oriW / 2)
+  console.log('x_left_threeshold: ' + x_left_threeshold)
+  console.log('x_right_threeshold: ' + x_right_threeshold)
```
thyeggman, line 164: Remove logging before merging

server/routes/api/representatives.js
thyeggman: Also check the variable names as well - should be spelled the same way

server/routes/api/representatives.js
```diff
@@ -129,6 +130,72 @@ router.get('/:zipCode', async (req, res) => {
...
+  // args
+  console.log(photo_cropping_object)
+  let x = photo_cropping_object.x
+  let y = photo_cropping_object.y
+  let oriH = photo_cropping_object.oriHeight
+  let oriW = photo_cropping_object.origWidth
+
+  // 1. calculate threeshold for the x space
+  // we check if the coordinate starts on the left side of the image (the first half of the left side)
+  let x_left_threeshold = oriW / 2 / 2
```
thyeggman, line 160:
```suggestion
  let x_left_threeshold = oriW / 4
```

server/routes/api/representatives.js
```diff
@@ -129,6 +130,72 @@ router.get('/:zipCode', async (req, res) => {
...
+  // we check if the coordinate starts on the right side of the image (the first half of the right side) and we reduce a margin of 5% to be flexible
+  let x_right_threeshold = oriW / 2 - (5 / 100) * (oriW / 2)
```
thyeggman, line 162: Extract the `(5 / 100)` into a variable so the margin can be adjusted for every calculation at the same time.  Also 0.05 would be a be a bit more readable for me, but that's just personal preference :)

server/routes/api/representatives.js
```diff
@@ -129,6 +130,72 @@ router.get('/:zipCode', async (req, res) => {
...
+  console.log('x_left_threeshold: ' + x_left_threeshold)
+  console.log('x_right_threeshold: ' + x_right_threeshold)
+  // 2. calculate threeshold for the y space
+  // we check if the coordinate starts on the top side of the image (the first half of the top side)
+  let y_top_threeshold = oriH / 2 / 2
```
thyeggman, line 167:
```suggestion
  let y_top_threeshold = oriH / 4
```

server/routes/api/representatives.js
```diff
@@ -129,6 +130,72 @@ router.get('/:zipCode', async (req, res) => {
...
+      ? 'left'
+      : x >= x_right_threeshold
+      ? 'right'
+      : 'center'
+  let y_value =
+    y <= y_top_threeshold
+      ? 'top'
+      : y >= y_bottom_threeshold
+      ? 'bottom'
+      : 'center'
```
thyeggman, line 185: To improve readability, don't chain the ternary operations together.  I would suggest assigning the second part of each ternary operation to its own variable, and then put the resulting value into the assignments here.

package.json
```diff
@@ -51,6 +51,7 @@
     "@stripe/stripe-js": "^1.35.0",
     "@vue-stripe/vue-stripe": "^4.4.2",
     "axios": "^0.27.2",
+    "axios-cache-interceptor": "^0.10.7",
```
thyeggman, line 54: This is probably a leftover from your testing for the cache implementation, it should not be included in this PR yet, since it's unrelated 😉  You should also remove the changes to package-lock.json

ahn-nath: I will keep this "relatedness" priority in mind. Thanks.

server/routes/api/representatives.js
ahn-nath: I do not have a preference about the 0.05. I agree about the value being more flexible with a variable.

server/routes/api/representatives.js
ahn-nath: I am not sure if I understand the suggestion. I understand the "if-else-if-else" block can cause confusion, but it is hard for me to picture the second part being alone (the else-if block) as another ternary operation. Nevertheless, I will do my best to improve readability.

ahn-nath: I have used if-else blocks to make the code easier to read.

ahn-nath:
@thyeggman 

I appreciate your comments and the time you invested on this PR; they were helpful.

paramsiddharth (changes requested):
@ahn-nath Hi, Nath. I think the change in the `package-lock.json` is accidental and shouldn't be present here. Could you please revert it to the one in the `main` branch? You can see the changes [here](https://github.com/ProgramEquity/amplify/pull/367/files) and realize that it mentions the `axios-cache-interceptor`. Maybe you installed it and then uninstalled it? Once you revert to the original `package-lock.json`, do an `npm ci` and it should fix everything. Let me know if you need my help.

ahn-nath:
> @ahn-nath Hi, Nath. I think the change in the `package-lock.json` is accidental and shouldn't be present here. Could you please revert it to the one in the `main` branch? You can see the changes [here](https://github.com/ProgramEquity/amplify/pull/367/files) and realize that it mentions the `axios-cache-interceptor`. Maybe you installed it and then uninstalled it? Once you revert to the original `package-lock.json`, do an `npm ci` and it should fix everything. Let me know if you need my help.

I will update it soon and commit the fix. Thanks!

server/routes/api/representatives.js
thyeggman: Sorry, I should have been clearer.  I just meant something like:

```js
  let x_right = x >= x_right_threshold ? 'right' : 'center'
  let x_value = x <= x_left_threshold ? 'left' : x_right
```

I think the code you replaced it with is also more readable.  Up to you if you want to switch back to using the ternary pattern, thanks for considering readability anyways!

ahn-nath: I agree this alternative is more readable, and I tried to prevent it from being verbose. Thanks. I will keep this version.

ahn-nath (commented):
@paramsiddharth 
I have reviewed the relevant files with a `git checkout origin/main package-lock.json` command.

thyeggman (dismissed):

paramsiddharth (dismissed):
@ahn-nath Good job, as always, Nath. Just tested it out locally. It works amazing! I appreciate how you described the design of `getPhotoCroppingValues` in the PR description. The function is really well-written and functioning. We don't need to worry about the `null` values — If the people at Cicero weren't able to figure those out for us, we shouldn't be concerned either, haha!

Despite it being a low-priority issue, you've done a good job with the research and quality, once again. Let's wait on Sam's PR (#366) to be merged to minimize the conflicts. Meanwhile, you can work on implementing #294, because you've already done the research and we have a clear idea that we're moving forward with the cache interceptor library. Keep it up!

sammychinedu2ky:
Hi @ahn-nath , the logic here is pretty cool 😅

thyeggman (approved):

PR #391, opened 2022-11-04
changed key to be unique

sammychinedu2ky:
I changed the key from the social media type to the iteration index as there are cases where the social media type appears twice which leads to warning errors during development.


paramsiddharth (approved):
Thank you for the fix, Sam! :) Looks good to me!

PR #396, opened 2022-11-08
Correct submission behavior for postal form

### LINKED ISSUE #293: Pressing enter in the postal code form refreshes page
# Pressing enter in the postal code form refreshes page

image

After writing the postal code, it is user-tendency to press enter and expect the form to submit itself.

However, it doesn't happen right now, because the submit behaviour has only been attached to the button
and not the form. The submission behaviour of the form needs to be modified.

## Refer
- [SearchReps.vue: Form](https://github.com/ProgramEquity/amplify/blob/main/src/components/SearchReps.vue#L9-L16)
- [SearchReps.vue: Submission Behaviour](https://github.com/ProgramEquity/amplify/blob/main/src/components/SearchReps.vue#L73-L81)

## Possible Solution
- The `on-submit` property of the `v-form` needs to recreate the `on-click` behaviour exhibited by the submit button.
- For event `e`, `e.preventDefault()` needs to be called to prevent the default form submission behaviour i.e. Making a request to the current route with the data, which causes the page refresh.

tdo95:
For the postal form submission we wanted to:
- recreate the `on-click` behaviour exhibited by the submit button
-  prevent the default form submission behaviour, which refreshes the page

### Solution
I used the `.prevent` [event modifier](https://vuejs.org/guide/essentials/event-handling.html#event-modifiers) to prevent the default submission behavior so that pressing enter no longer refreshes the page, but instead executes the CreateRepList function to display the representatives.

ezgif com-gif-maker

--
Resolves #293

### DIFF
--- a/src/components/SearchReps.vue
+++ b/src/components/SearchReps.vue
@@ -7,7 +7,7 @@
             <v-card-text>
               <v-subheader class="pa-0"> Where do you live? </v-subheader>
 
-              <v-form ref="form">
+              <v-form @submit.prevent="CreateRepList()" ref="form">
                 <v-text-field
                   v-model="postalCode"
                   label="Postal Code"


paramsiddharth (approved):
@tdo95 Good job, Tee! :) Thank you for the much needed fix!

Issue #399, opened 2022-11-11
Support queries by postal address


thyeggman:
Currently, we only support queries by ZIP code.  Zip codes do not necessarily correlate with legislative districts.  By enabling searches by address, the Cicero API will return only the results for the legislative districts for that point.  This will give more granularity to search results for local representatives, and there won't be representatives presented outside the sender's legislative area(s).  Since we have caching already, this should mitigate the effects of increased cost due to additional granularity in results.

- [ ] Add another API to the app which accepts an address in its body
- [ ] Feed the address information to the [cicero address API](https://cicero.azavea.com/docs/#address)
- [ ] Update the Amplify App UI to call the new endopint if the search term looks like an address

These should be treated as separate tasks.  However, once the endpoint name is determined, work can start on the front end side to update the UI to call the new API.

paramsiddharth:
> > * Query/Cache the Cicero API for the officials based on the district ID
> 
> @paramsiddharth Though I am using the location queries specified in the description of this issue, I am not using them with the legislative-district or nonlegislative-district endpoints. Which, according to my tests and involvement with the API, are the ones that return the district IDs, at least solely (the response specifically returns objects corresponding to the districts). The reason for not using these IDs is that you cannot search officials (referring to the `officials` endpoint) with the district_id alone, you need to include the location queries specified in [cicero.azavea.com/docs/#location-queries](https://cicero.azavea.com/docs/#location-queries). Without them, filtering by IDs will not work. The official resource works with location queries, including the street address. I have decided to directly implement it with it. Fetching the district ID by type (you have to make two calls for each type, in the best case scenario), along with the location query values on the bottom of the response (that the official resource also has), to then make another call to get the officials does not seem efficient for this case.
> 
> I don't think you need to update the steps in this reference, as it is clear to me now, but I am mentioning it because my implementation for this changed in accordance with my new understanding of the solution.

@ahn-nath Makes sense, Nath. I agree. The caching proposal I mentioned above is just to collect the ideas we discussed in our last meeting, so don't worry about firmly implementing it right away. It is something we plan to deal with once we are done with supporting address queries first, and I believe we definitely need a lot more planning than what we have done so far before even starting to implement the caching for the address queries. So you can leave it for now, and if you have already implemented some of the logic, you can comment it out and can resume working on it later. Meanwhile, we are just focusing on having the postal address query support with your ongoing draft PR.

@thyeggman What do you think about what Nath said above, Jacob?

[...]

PR #262, opened 2022-07-17
Update to Action complete issue #215

DietBepis1:
- Adds basic functionality to ActionComplete.vue
- Implements a carousel in CauseCarousel.vue
- Creates a generic CampaignCard component to reduce code repetition.

The purpose of the PR is to prep the completion page for adding persisted campaign information from before the Stripe redirect. This way, we can repopulate representative information and user information that is currently lost.

PR #402, opened 2022-11-17
Add another API call to the app which accepts an address in its body

ahn-nath:
> Enable searches by address in order for the Cicero API to return only the results for the legislative districts for that point. This will give more granularity to search results for local representatives, and there won't be representatives presented outside the sender's legislative area(s). Since we have caching already, this should mitigate the effects of increased cost due to additional granularity in results.

- [ ] Add another API call to the app which accepts an address in its body
- [ ]  Feed the address information to the [cicero address API](https://cicero.azavea.com/docs/#address)
- [ ]  Update the Amplify App UI to call the new endopint if the search term looks like an address
- [ ]  Follow up on the caching – we currently use the ZIP code as the key, but we'll need to think of different addresses may refer to the same set of reps

### DIFF
--- a/server/routes/api/representatives.js
+++ b/server/routes/api/representatives.js
@@ -129,6 +129,17 @@ router.get('/:zipCode', async (req, res) => {
   }
 })
 
+router.post('/districts', async (req, res) => {
+  const { address } = req.body
+
+  try {
+    console.log('address', address)
+  } catch (error) {
+    console.log(error)
+    res.status(500).send({ error: 'Whoops' })
+  }
+})
+
 function getOfficialSocialMediaPages(identifiers) {
   var social_media_pages = []
 
--- a/src/components/SearchReps.vue
+++ b/src/components/SearchReps.vue
@@ -183,14 +183,31 @@ export default {
         },
         async CreateRepList() {
             try {
-                const res = await axios.get(
+
+                // check postal code is valid with regex
+                let res = ''
+                let isPostalCodeValid =  /(^\d{5}$)|(^\d{5}-\d{4}$)/.test(this.postalCode);
+                if(!isPostalCodeValid) {
+                  res = await axios.get(
                     '/api/representatives/' + this.postalCode
                 )
-                this.isActive = false
+                }else{
+                  // check if street address is valid with a flexible regex. This validation is not perfect, but caches common cases
+                  // [a-zA-Z0-9\s]	Any single character in the range a-z or A-Z or 0-9 or whitespace
+                  // '-	Matches the characters - (case sensitive)
+                  // * 0 or more times
+                  // $  end of string
+                  let streetAddressValid = /^[a-zA-Z0-9\s,'-]*$/.test(this.postalCode);
+                  if(streetAddressValid){
+                  res = await axios.post(
+                    '/api/representatives/districts', {address: this.postalCode}
+                  )
+                  }
+                }
 
+                this.isActive = false
                 this.congressMembers = res.data
                 this.hasContent = true
-                // console.log(res.data)
                 this.listVisible = true
             } catch (e) {
                 console.error(e)


paramsiddharth:
@ahn-nath Good going, Nath. :) Keep us posted with your progress, and also, run the linter often to make sure the code remains formatted.

src/components/SearchReps.vue
```diff
@@ -196,9 +196,29 @@ export default {
...
                     '/api/representatives/' + this.postalCode
                 )
+
+                }else{
+                  // check if street address is valid with a flexible regex. This validation is not perfect, but caches common cases
+                  // [a-zA-Z0-9\s]	Any single character in the range a-z or A-Z or 0-9 or whitespace
+                  // '-	Matches the characters - (case sensitive)
+                  // * 0 or more times
+                  // $  end of string
+                  let streetAddressValid = /^[a-zA-Z0-9\s,'-]*$/.test(this.postalCode);
```
thyeggman, line 214: The name of the parameter should change from `postalCode` to something like `searchTerm`

When searching by address, the [API request body](https://cicero.azavea.com/docs/#address) is split up into multiple parts like `search_address`, `search_city`, etc.  There needs to be more complex parsing to split up the search term, or we need a new way to provide multiple separate inputs

ahn-nath: @thyeggman 
But the search_address parameter is enough to generate a successful and logical response that returns several districts. I have tried https://cicero.azavea.com/v3.1/legislative_district?search_loc=990+spring+garden+st+philadelphia+pa&format=json&key=405f15046940ce2a712b41c0f442b1ac2e57e3b9
. I feel that if the team wants it to include more parameters like search_city and even more than that, the input cannot be a simple input text, it should be stricter than this. Regex enough cannot or should not handle so many edge cases, I believe.

thyeggman: Ah, you're right, my bad!  For some reason my brain completely ignored the alternate `search_loc` parameter that can be used 😅

ahn-nath: No worries. I will update the rest of the code to fix the issues you found.

ahn-nath:
**Pending**
I still need to improve the Regex lines used to validate the street address.

src/components/SearchReps.vue
```diff
@@ -188,17 +175,43 @@ export default {
...
         },
         async CreateRepList() {
             try {
-                const res = await axios.get(
-                    '/api/representatives/' + this.postalCode
+
+                // check postal code is valid with regex
+                let res = ''
+                let isPostalCodeValid =  /(^\d{5}$)|(^\d{5}-\d{4}$)/.test(this.searchText);
+                let streetAddressValid = /^[a-zA-Z0-9\s,'-]*$/.test(this.searchText);
```
thyeggman, line 190: I know you mentioned wanting to improve this, this site allows you to search for common regex patterns: [link](https://regexlib.com/Search.aspx?k=us+street+address&c=-1&m=-1&ps=20)

I linked a search for US street address, but I also found results for postal code.  I wouldn't worry about getting something that matches everything exhaustively, but rather something that works "well enough" - the cicero API just won't return results if it doesn't find any for the given address, so we can put a note on the page for the user to check the address they entered and look for typos or something similar

ahn-nath: Thanks!

ahn-nath:
@thyeggman @paramsiddharth 
Normally, I just re-request review from one mentor/dev lead, but since you are both assigned to this task, I am tagging you both. 

Looking forward to your comments.

paramsiddharth (changes requested):
@ahn-nath The work looks good to me, Nath. :) There are two things I would like to point out:
1. I don't think the intention of #399 was to completely replace the zip-code queries by addresses. Instead, I think it meant _adding_ the ability to also be able to support queries by address while still keeping the ability to search by zip-code intact (@thyeggman Correct me if I am wrong)

   What that implies is, both the back-end endpoints must be preserved. The front-end should also contain two query options to choose from i.e. Either use zip-code or use full address.
2. After merging #262, there seem to be new conflicts in your PR (in `SearchReps.vue`). Please try to resolve them and update the PR.

That's it, and after that, I think your PR will be ready to merge! :)

ahn-nath:
> @ahn-nath The work looks good to me, Nath. :) There are two things I would like to point out:
> 
> 1. I don't think the intention of #399 was to completely replace the zip-code queries by addresses. Instead, I think it meant _adding_ the ability to also be able to support queries by address while still keeping the ability to search by zip-code intact (@thyeggman Correct me if I am wrong)
>    What that implies is, both the back-end endpoints must be preserved. The front-end should also contain two query options to choose from i.e. Either use zip-code or use full address.
> 2. After merging #262, there seem to be new conflicts in your PR (in `SearchReps.vue`). Please try to resolve them and update the PR.
> 
> That's it, and after that, I think your PR will be ready to merge! :)

Hey, @paramsiddharth. Thanks for your feedback. 

1. I do not think having two endpoints makes sense, since the logic is exactly the same, we are using the same Cicero API call, and only one parameter changes. I initially intended to have two endpoints, but after testing the different options and actually implementing it as a draft, I can now see this issue as **_an extension of the functionality that we already have_**. We are supporting street addresses to provide more granularity as opposed to replacing or using a different logic to get representatives. As a matter of fact, we are just using/adding a new location query. They are not exclusive to each other in any sense, they can work together as a group, just like the filters do. 

What I suggest, instead, is that we have two different endpoints, one for the ZIP code, as we already have, and one for the street address input. And then, we simply send them both as parameters after checking that they are valid and not empty. That is, if the person only inputs a valid ZIP code and not a street address, we simply do not add the address. The same goes when the opposite happens.

2. I will fix this tomorrow.

paramsiddharth:
> 1. I do not think having two endpoints makes sense, since the logic is exactly the same, we are using the same Cicero API call, and only one parameter changes. I initially intended to have two endpoints, but after testing the different options and actually implementing it as a draft, I can now see this issue as **_an extension of the functionality that we already have_**. We are supporting street addresses to provide more granularity as opposed to replacing or using a different logic to get representatives. As a matter of fact, we are just using/adding a new location query. They are not exclusive to each other in any sense, they can work together as a group, just like the filters do.
> 
>    What I suggest, instead, is that we have two different text inputs, one for the ZIP code, as we already have, and one for the street address input. And then, we simply send them both as parameters after checking that they are valid and not empty. That is, if the person only inputs a valid ZIP code and not a street address, we simply do not add the address. The same goes when the opposite happens.

@thyeggman What do you think about this, Jacob?

paramsiddharth:
> Manisha: Make the full address query **mandatory** for the local and school filters. _[Can be discussed more]_

src/components/SearchReps.vue
```diff
@@ -188,18 +175,27 @@ export default {
...
-                const res = await axios.get(
-                    '/api/representatives/' + this.postalCode
+                // check postal code is valid with regex
+                let res = ''
+                let isPostalCodeValid =  /(^\d{5}$)|(^\d{5}-\d{4}$)/.test(this.searchText);
+                let streetAddressValid = /^(\d{3,})\s?(\w{0,5})\s([a-zA-Z]{2,30})\s([a-zA-Z]{2,15})\.?\s?(\w{0,5})$/.test(this.searchText);
+                if(isPostalCodeValid || streetAddressValid) {
+                  console.log('valid input to get representatives')
+                  res = await axios.get(
+                    '/api/representatives/' + this.searchText
```
thyeggman, line 195: I'm not sure if if this will work as written.  The search text should not be passed as part of the route, it should be a request parameter.  It was a bad pattern to use with ZIP codes as well, but it worked because none of the characters in a ZIP code need to be encoded in a URL.  Addresses can have spaces and other blank space in them, which needs to be encoded in a URL (e.g. a space becomes `%20`).

So you'll need to change the call to get, and change the route to take `searchText` as either a parameter in the body of the request or a URL query parameter (e.g. `&searchText=<address>`).  The former is a better option, as encoding just adds extra complexity and the address could be quite long.

server/routes/api/representatives.js
```diff
@@ -50,15 +50,32 @@ const ALLOWED_JURISDICTION_FILTERS = Object.keys(JURISDICTION_FILTER_MAP)
 // Endpoints
 
 // Get
-router.get('/:zipCode', async (req, res) => {
-  const { zipCode } = req.params
+router.get('/:searchText', async (req, res) => {
```
thyeggman, line 53: This should be changed so the search address in not part of the route, but is provided as a request parameter in the body.

server/routes/api/representatives.js
```diff
@@ -50,15 +50,32 @@ const ALLOWED_JURISDICTION_FILTERS = Object.keys(JURISDICTION_FILTER_MAP)
...
+  let isValidZIPcode = /^\d{5}(-\d{4})?$/.test(searchText)
+  /* check if valid street address
+     ^(\d{3,})\s? the start of the line can have 3 or more digits followed by a space
+     (\w{0,})\s the next part can have 2 or more letters followed by a space
+     ([a-zA-Z]{2,30})\s the next part must be an alphanetical string with 2-30 characters followed by a space
+     ([a-zA-Z]{2,15})\.?\s?  the next part must be an alphabetical string with 2-15 characters followed by a any digit (optionally) and/or space (optionally)
+     (\w{0,5})$ the end of the line can have 0-5 letters
+  */
+  let isValidAddress =
+    /^(\d{3,})\s?(\w{0,5})\s([a-zA-Z]{2,30})\s([a-zA-Z]{2,15})\.?\s?(\w{0,5})$/.test(
```
thyeggman, line 70: The regex is probably more robust if it requires a space (`+`) instead of allowing one or zero (`?`)

```suggestion
    /^(\d{3,})\s+(\w{0,5})\s+([a-zA-Z]{2,30})\s+([a-zA-Z]{2,15})\.?\s+(\w{0,5})$/.test(
```

server/routes/api/representatives.js
```diff
@@ -50,15 +50,32 @@ const ALLOWED_JURISDICTION_FILTERS = Object.keys(JURISDICTION_FILTER_MAP)
...
+  /* check if valid ZIP code 
+    ^\d{5} - the start of the line has 5 digits
+    (-\d{4})?$ the end of the line has 4 digits preceded by a dash (optionally)
+  */
+  let isValidZIPcode = /^\d{5}(-\d{4})?$/.test(searchText)
+  /* check if valid street address
+     ^(\d{3,})\s? the start of the line can have 3 or more digits followed by a space
+     (\w{0,})\s the next part can have 2 or more letters followed by a space
+     ([a-zA-Z]{2,30})\s the next part must be an alphanetical string with 2-30 characters followed by a space
+     ([a-zA-Z]{2,15})\.?\s?  the next part must be an alphabetical string with 2-15 characters followed by a any digit (optionally) and/or space (optionally)
```
thyeggman, line 66: This does not accept a digit as written, it accepts exactly `.` because the regex includes `\.` to escape the period.  Matching zero or one digit would be `\d?`. I'm not sure why this part is allowed to end in a single digit, though?

server/routes/api/representatives.js
```diff
@@ -50,15 +50,32 @@ const ALLOWED_JURISDICTION_FILTERS = Object.keys(JURISDICTION_FILTER_MAP)
...
 
-  if (!zipCode.match(/^\d{5}(-\d{4})?$/)) {
+  /* check if valid ZIP code 
+    ^\d{5} - the start of the line has 5 digits
+    (-\d{4})?$ the end of the line has 4 digits preceded by a dash (optionally)
+  */
+  let isValidZIPcode = /^\d{5}(-\d{4})?$/.test(searchText)
+  /* check if valid street address
+     ^(\d{3,})\s? the start of the line can have 3 or more digits followed by a space
+     (\w{0,})\s the next part can have 2 or more letters followed by a space
```
thyeggman, line 64: This part will not match addresses like `123 45th street` - Addresses with ordinal parts (e.g. 1st, 2nd, 3rd) would be invalid.  I would recommend something similar to `(\d+\w{2})|((\w{2,})\s?)+` (I didn't test this) to match ordinal addresses and also street names with multiple words

thyeggman (changes requested):
There are several issues with the regex - but again, I'm not sure it's necessary, since the Cicero API does its own address validation to be able to search for representatives.  For our purposes, it's probably sufficient to check if it looks like a zip code, and if it doesn't, then send it to cicero as a `search_loc` parameter.  Then if the result is empty, we render a message on the page for the user to check the address.

@paramsiddharth There was a bit of a misunderstanding this morning - this implementation does not remove the existing zip-code functionality.  The problem is that we are misusing the API route by sending the search information as part of the route instead of as a request parameter.  So it _looked_ like the route was changing (and it is, as written), but it would have been used the same way.  We should just remove the query text from the route to solve this issue, and we can maintain a single endpoint still.

ahn-nath:
I would like to take this chance to reference the regex I used for the street address:
**Line:**  ^(\d{3,})\s?(\w{0,5})\s([a-zA-Z]{2,30})\s([a-zA-Z]{2,15})\.?\s?(\w{0,5})$
**Source:** https://regexlib.com/REDetails.aspx?regexp_id=2420
**Author:** The author is under the username "NoBug" on the source website. 
**Reason:** I need to develop my understanding of ReGex and have used a suggested website to find more sophisticated solutions.

ahn-nath:
> There are several issues with the regex - but again, I'm not sure it's necessary, since the Cicero API does its own address validation to be able to search for representatives. For our purposes, it's probably sufficient to check if it looks like a zip code, and if it doesn't, then send it to cicero as a `search_loc` parameter. Then if the result is empty, we render a message on the page for the user to check the address.
> 

As you mentioned, I do not think we can rely on a regular expression to guarantee a valid street address each time. The best way to validate it is by using another API, and this is already done by the Cicero API. I am still using regex to at least prevent the app from triggering a meaningless request each time and wasting credits unnecessarily. That is, there is no point in validating a dot (.) if the system already “knows” it is obviously invalid input and do not even meet the most basic requirements. Besides that, being too strict with the format at the time we receive an input can also be an issue, because we will still use a flexible locator.

ahn-nath:
> @paramsiddharth There was a bit of a misunderstanding this morning - this implementation does not remove the existing zip-code functionality. The problem is that we are misusing the API route by sending the search information as part of the route instead of as a request parameter. So it _looked_ like the route was changing (and it is, as written), but it would have been used the same way. We should just remove the query text from the route to solve this issue, and we can maintain a single endpoint still.

> Manisha: Make the full address query mandatory for the local and school filters. [Can be discussed more]

If the street address needs to be mandatory for some filters, in addition to removing the query text from the route, do not you think that we should add a separate input for the street address specifically? So that it is easier to validate and generates less ambiguity. All things considered, we can still search with multiple address components and send a query that contains both the `search_postal` and the `search_address`. At the end of the day, they both use a geocoding service.

ahn-nath:
> > Manisha: Make the full address query **mandatory** for the local and school filters. _[Can be discussed more]_

@paramsiddharth 
Hi, Param. When you have them, please specify the exact details of this requirement for me to work on it. Thanks.

thyeggman:
> I am still using regex to at least prevent the app from triggering a meaningless request each time and wasting credits unnecessarily.

That's a good thought, I would just say don't let "perfect" be the enemy of "good" in this case - get something that validates basic format and don't worry that it doesn't satisfy the exact formatting of an official address

paramsiddharth:
> @paramsiddharth There was a bit of a misunderstanding this morning - this implementation does not remove the existing zip-code functionality. The problem is that we are misusing the API route by sending the search information as part of the route instead of as a request parameter. So, it _looked_ like the route was changing (and it is, as written), but it would have been used the same way. We should just remove the query text from the route to solve this issue, and we can maintain a single endpoint still.

@thyeggman I agree with you! I still think we can use the zip code as a route if we make sure we encode it properly in the front-end, but for something like a complete address, I definitely believe we should use the request parameter (or the request body if it is a `POST`/`PUT` request). Retaining the same end-point sounds reasonable.

> @thyeggman, @paramsiddharth If the street address needs to be mandatory for some filters, in addition to removing the query text from the route, do not you think that we should add a separate input for the street address specifically? So that it is easier to validate and generates less ambiguity. All things considered, we can still search with multiple address components and send a query that contains both the `search_postal` and the `search_address`. At the end of the day, they both use a geocoding service.

@ahn-nath I agree, Nath. That was what I was planning, too. I think it would be nice if we could do a 1:1 to discuss this through. There's a lot of ambiguity in text-based communication, and we are also skipping the pair hours this week due to holidays in the US.

> > > Manisha: Make the full address query **mandatory** for the local and school filters. _[Can be discussed more]_
> 
> @paramsiddharth Hi, Param. When you have them, please specify the exact details of this requirement for me to work on it. Thanks.

I'll think about the requirements properly again, and then I will explain to you what they are.

ahn-nath:
> @thyeggman I agree with you! I still think we can use the zip code as a route if we make sure we encode it properly in the front-end, but for something like a complete address, I definitely believe we should use the request parameter (or the request body if it is a `POST`/`PUT` request). Retaining the same end-point sounds reasonable.

I will send my update/commit with the endpoint without the query text. Since the search_postal and search_address are both query locators, semantically, I feel it is more consistent to have them all being part of the request body. If the search address and the postal code can both be optional (for example, I can send the street address and not necessarily the postal code), then we would be sending an empty string, just to make sure we match the endpoint. With both being part of the request body, we can just check for their presence. Similarly, if we decide to add more query locators, such as the latitude and longitude, it will make sense to have them all as part of the request body, instead of making an exception with the ZIP code or having to choose which gets to be formatted. 

> @ahn-nath I agree, Nath. That was what I was planning, too. I think it would be nice if we could do a 1:1 to discuss this through. There's a lot of ambiguity in text-based communication, and we are also skipping the pair hours this week due to holidays in the US.
I agree. I will contact you on Slack to discuss this.

> I'll think about the requirements properly again, and then I will explain to you what they are.
Awesome. Then I will move on with the rest of the requirements.

server/routes/api/representatives.js
ahn-nath: @thyeggman 
A street address can have abbreviations like these: 101 W Main St. N. The dot makes sense to me. And actually, it accepts any single character after the limit. I have tested it with https://rubular.com/. For example, if you change the limit to 0-2 or 1 alphabetic characters and any single character after it` ^(\d{3,})\s?(\w{0,5})\s([a-zA-Z]{2,30})\s([a-zA-Z]{0,2}).?\s?(\w{0,5})$`, the results are as follows:

- 101 W Main S21 N will not have matches or be valid
- 101 W Main St1 N will be valid (any combination of two alphabetic characters plus any character will work, not only the dot).

server/routes/api/representatives.js
ahn-nath: @thyeggman 
I do not agree with this. Some sections should be optional, and this regex should also target several cases. I am not as familiar with US addresses, but as far as I know, these are all valid:
- 101 Main St N
- 101 W Main St 
- 123 45th street (you used this as an example for another suggestion).

But they would not work with the new suggestion.

paramsiddharth:
Considering the fact that we are using the address/zip-code _only_ to query the _representatives_ of that region, we can minimize the address-validation. :)

server/routes/api/representatives.js
ahn-nath: I have updated this to have it as a URL query parameter.

src/components/SearchReps.vue
ahn-nath: Hello, @thyeggman. I have updated the input text to be a URL query parameter when it is a street address (or the string is valid for a street address with the initial regex check), keeping in mind that in the near future, @paramsiddharth and @manishapriya94 seem to suggest we can have two input text fields and therefore keep two parameters. 


Second, @paramsiddharth and you seem to differ on whether the postal code should be a URL query parameter or not. If we change them to be part of the body, I need to make a POST request instead and add each parameter to the body of the request. It is important for me that you both come to a single conclusion, as I have already stated my position regarding it.

ahn-nath:
**Your comments:** 
> Considering the fact that we are using the address/zip-code _only_ to query the _representatives_ of that region, we can minimize the address-validation. :)

**Jacobs' comments:**
> I'm not sure if if this will work as written. The search text should not be passed as part of the route, it should be a request parameter. It was a bad pattern to use with ZIP codes as well, but it worked because none of the characters in a ZIP code need to be encoded in a URL. Addresses can have spaces and other blank space in them, which needs to be encoded in a URL (e.g. a space becomes %20).

> So you'll need to change the call to get, and change the route to take searchText as either a parameter in the body of the request or a URL query parameter (e.g. &searchText=<address>). The former is a better option, as encoding just adds extra complexity and the address could be quite long.

Hello, @paramsiddharth. I just re-read Jacob's suggestion in relation to still keeping the zip code as part of the URL, and even further, as keeping the route a GET call and I would like you both to come to one single suggestion. As you are both my mentors and I have stated my preference already.

paramsiddharth:
> Hello, @thyeggman. I have updated the input text to be a URL query parameter when it is a street address (or the string is valid for a street address with the initial regex check), keeping in mind that in the near future, @paramsiddharth and @manishapriya94 seem to suggest we can have two input text fields and therefore keep two parameters.
> 
> Second, @paramsiddharth and you seem to differ on whether the postal code should be a URL query parameter or not. If we change them to be part of the body, I need to make a POST request instead and add each parameter to the body of the request. It is important for me that you both come to a single conclusion, as I have already stated my position regarding it.

> **Your comments:**
> 
> > Considering the fact that we are using the address/zip-code _only_ to query the _representatives_ of that region, we can minimize the address-validation. :)
> 
> **Jacobs' comments:**
> 
> > I'm not sure if if this will work as written. The search text should not be passed as part of the route, it should be a request parameter. It was a bad pattern to use with ZIP codes as well, but it worked because none of the characters in a ZIP code need to be encoded in a URL. Addresses can have spaces and other blank space in them, which needs to be encoded in a URL (e.g. a space becomes %20).
> 
> > So you'll need to change the call to get, and change the route to take searchText as either a parameter in the body of the request or a URL query parameter (e.g. &searchText=). The former is a better option, as encoding just adds extra complexity and the address could be quite long.
> 
> Hello, @paramsiddharth. I just re-read Jacob's suggestion in relation to still keeping the zip code as part of the URL, and even further, as keeping the route a GET call and I would like you both to come to one single suggestion. As you are both my mentors and I have stated my preference already.

@ahn-nath Hi, Nath. Oh, I just wanted to point out that if the payload gets too expensive in terms of size and complexity, we can consider using a `POST` endpoint. It was not a preference. :) Right now it seems in order, and the fact that the endpoint is only meant to fetch (and not modify) data, a `GET` endpoint seems semantically correct to use here. I think you should go ahead with the `GET` endpoint itself. :)

server/routes/api/representatives.js
ahn-nath: Hello, this part will not accept a string like 123 45th street, because it expects 2 or more alphabetical characters for the second group. Addresses with ordinal parts (e.g., 1st, 2nd, 3rd) would be valid if they had a third group with additional alphabetical characters.  The w{start, finish} already implies any word character (letter, number, underscore). That is why I somehow feel (\d+\w{2})|((\w{2,})\s?)+ could be redundant as we are already implying we can accept any word character (numbers or alphabetical characters as part of the second group). I have updated the third group to be optional and take strings that do not have two additional words after the second group, let me know what you think!

ahn-nath: @thyeggman 
**The update:**
`(\d{3,})\s?(\w{0,5})\s([a-zA-Z]{2,30})\s?([a-zA-Z]{2,15})\.?\s?(\w{0,5})$`

ahn-nath:
@thyeggman 
I have re-requested a review after marking items as solved, as I have not received an update in 7 days. 

@paramsiddharth 
Please remember to send the Loom video and/or instructions on how the new requirement would be implemented.

Thanks.

paramsiddharth:
> @thyeggman I have re-requested a review after marking items as solved, as I have not received an update in 7 days.
> 
> @paramsiddharth Please remember to send the Loom video and/or instructions on how the new requirement would be implemented.
> 
> Thanks.

@ahn-nath Hi Nath, we will get back to you very soon. Jacob is off this week, so we will talk more about it whenever possible. About the issue, there's a lot of discussion to be made and doing it asynchronously doesn't seem to work anymore, so let's wait till we can all discuss this together over a meeting sometime. Till then, you can take a break, too. :)

server/routes/api/representatives.js
ahn-nath: I have made a mistake and fixed it by removing the escaping character.

thyeggman:
I don't mind being pinged - it helps me keep track of what I need to look at when I get back!  I'm back as of today, I'll look at the changes

src/components/SearchReps.vue
```diff
@@ -195,23 +182,47 @@ export default {
...
+                let params = {streetAddress: this.searchText}
+                let isPostalCodeValid =  /(^\d{5}$)|(^\d{5}-\d{4}$)/.test(this.searchText);
+                let streetAddressValid = /^(\d{3,})\s?(\w{0,5})\s([a-zA-Z]{2,30})\s([a-zA-Z]{2,15})\.?\s?(\w{0,5})$/.test(this.searchText);
+
+
+                if(isPostalCodeValid) {
+                 // console.log('valid input to get representatives with postal code')
+                  res = await axios.get(
+                    '/api/representatives/' + this.searchText
+                )
```
thyeggman, line 206:
```suggestion
                  )
```

src/components/SearchReps.vue
```diff
@@ -195,23 +182,47 @@ export default {
...
+                }
+
+                else if(streetAddressValid) {
+                  // console.log('valid input to get representatives with street address')
+                  res = await axios.get(
+                    '/api/representatives/' + ' ', {
+                        params
+                    }
                 )
```
thyeggman, line 215:
```suggestion
                  )
```

src/components/SearchReps.vue
```diff
@@ -195,23 +182,47 @@ export default {
...
+                    '/api/representatives/' + this.searchText
+                )
+                }
+
+                else if(streetAddressValid) {
+                  // console.log('valid input to get representatives with street address')
+                  res = await axios.get(
+                    '/api/representatives/' + ' ', {
+                        params
+                    }
```
thyeggman, line 214: Just formatting - no need to add constants together instead of just combining them

```suggestion
                    '/api/representatives/ ', { params }
```

src/components/SearchReps.vue
```diff
@@ -226,8 +237,10 @@ export default {
                         params.filter = this.currentFilter
                     }
 
+                    // testing
```
thyeggman, line 240: Not sure what this is still doing here?

thyeggman (commented):
Just a few formatting nitpicks, thanks for updating the PR!

thyeggman (dismissed):

ahn-nath:
> I don't mind being pinged - it helps me keep track of what I need to look at when I get back! I'm back as of today, I'll look at the changes

Hello, Jacob. I was not aware/informed you were off. Glad my message did not disturb your space at the time. I have updated the code to fix the issue you pointed out. Thanks!

paramsiddharth (approved):

paramsiddharth:
@ahn-nath Good work, Nath! I made some related and unrelated changes and am merging the PR now.

PR #428, opened 2022-12-15
added customization

sammychinedu2ky:
Pull request for this issue https://github.com/ProgramEquity/amplify/issues/370
image

@paramsiddharth

PR #433, opened 2022-12-20
Sogorea te onboard

DietBepis1:
-Implemented a generic 'single-campaign mode'.
-Numerous bug and formatting fixes.


paramsiddharth (commented):
@DietBepis1 The PR looks good to me! I'll test it out once locally and let you know if there's any bug I notice.

One thing though, do you think the file should be named `sogorea-te.less`? Since we are merging it to `main`, can we try to name it to something more generic, like `current-campaign.less`?

DietBepis1:
> @DietBepis1 The PR looks good to me! I'll test it out once locally and let you know if there's any bug I notice.
> 
> One thing though, do you think the file should be named `sogorea-te.less`? Since we are merging it to `main`, can we try to name it to something more generic, like `current-campaign.less`?

Yep, totally agree. That file was an artifact from the Sogorea Te' event and was deleted in my most recent commit. There is a method for globally changing styles now, which is documented [here.](https://github.com/ProgramEquity/amplify/wiki/Single-Campaign-Mode#overriding-brand-colors)

paramsiddharth (approved):
@DietBepis1 It looks great to me! I think we're ready to merge it, and should do so without delay so that we can move on to resolving conflicts of and merging #428 by Sam (@sammychinedu2ky). There are, however, two points I'd like to mention:
* The error that prevents the letter from updating even after a representative is selected seems to still be happening. I'm not sure if it is because I'm missing the correct Lob test API key or if the error has been in the application since back. Let's discuss that on a personal 1:1 some time.

  select rep letter not updating
* Although you have clearly updated the social media URLs in the code, the links in the bottom of the page seem not to work right now, but we can defer that to another issue.

sammychinedu2ky:
> @DietBepis1 It looks great to me! I think we're ready to merge it, and should do so without delay so that we can move on to resolving conflicts of and merging #428 by Sam (@sammychinedu2ky). There are, however, two points I'd like to mention:
> 
> * The error that prevents the letter from updating even after a representative is selected seems to still be happening. I'm not sure if it is because I'm missing the correct Lob test API key or if the error has been in the application since back. Let's discuss that on a personal 1:1 some time.
>   select rep letter not updating
>         
>           [
>             
>               select rep letter not updating
>             
>           ](https://user-images.githubusercontent.com/30315706/209785484-6ed4d0a1-15ac-4411-a55d-e4dddcfc7d78.gif)
>           
>           
>             
>               
>                 
>               
>               
>                 
>                 
>               
>             
>             [
>               
>                 
>               
>             ](https://user-images.githubusercontent.com/30315706/209785484-6ed4d0a1-15ac-4411-a55d-e4dddcfc7d78.gif)
> * Although you have clearly updated the social media URLs in the code, the links in the bottom of the page seem not to work right now, but we can defer that to another issue.

Seems to be an old issue. I think creating  different issue for this would be okay

DietBepis1:
@paramsiddharth I am having that issue with letter load as well, specifically with the Sogorea Te' campaign. It has to do with how our db is set up. If you hardcode a different campaign id into Vuex's store, you should be able to load another letter. If you look in the browser console, I believe it's the getLetter endpoint that doesn't recognize that campaign id. But definitely double check me in case I have made some other mistake.

Expanding the data structures of Amplify is at the top of my list for the new year.

I will have to look at the social media links since those were working for me before I pushed my latest commit.

paramsiddharth:
@DietBepis1 Do you suggest we merge this PR and defer the issues to another? Sam's PR has been hanging around for too long so I was wondering if it is a good idea to merge both PRs and resolve the conflicts and build up from there.

paramsiddharth:
I just generated the 2 new issues we needed:
- #445 
- #446 

@DietBepis1 Merging this PR. :)

Issue #467, opened 2023-02-01
Documentation request: Cicero/Lob API key setup for development


RyanJamesCaldwell:
## Expected Behavior

There should be some setup instructions under the [Project Setup](https://github.com/ProgramEquity/amplify#project-setup) header in the README that explains how to request, set, and use the Cicero and Lob API keys for use in development.

## Context (Environment)

Local/Codespaces development

manishapriya94:
<img width="243" alt="Screen Shot 2023-02-01 at 12 57 48 PM" src="https://user-images.githubusercontent.com/9143339/216161734-5a15c400-17c4-41eb-b516-63ae2f09ec39.png">

paramsiddharth:
Linking to #431.

Issue #431, opened 2022-12-18
[Docs] Add documentation for local development


paramsiddharth:
# Add documentation for local development
The documentation and wiki currently lack sufficient content on setting up local development environments on Linux, Macintosh, and especially Windows. We have come across several bugs in Codespaces recently, which makes it not the best choice for development.

Although the `bootstrap` script makes things convenient for Homebrew users, the convenience we desire doesn't necessarily have to be a bootstrapping script but giving the contributors the ability to understand what dependencies and configuration they need to have to develop the application locally. Hence, we need to have documentation on setting up local development environments.

## Notes
- We need multiple guides each for Windows, Linux (guides for multiple distros would be great), and Macintosh.
- We need to take into account the infeasibility of having to figure out configurations for the environment variables, `.env` and the [`auth_config.json`](https://github.com/ProgramEquity/amplify/blob/main/auth_config.json) on one's own, and therefore should carefully document each of the configurations needed (including the optional ones) in great detail.
> - To add more.

paramsiddharth:
- [ ] #467

smgraywood:
Started here for MacOS: https://github.com/ProgramEquity/amplify/wiki/Local-Setup

paramsiddharth:
@smgraywood Good job, Sarah! Here's a suggestion: Try to use block-level codeblocks instead of inline codeblocks.

For example,

Instead of 
run `$ echo 123`
```
run `$echo 123`
```

Use
```shell
echo 123
```
````
```shell
echo 123
```
````

smgraywood:
> @smgraywood Good job, Sarah! Here's a suggestion: Try to use block-level codeblocks instead of inline codeblocks.
> 
> For example,
> 
> Instead of run `$ echo 123`
> 
> ```
> run `$echo 123`
> ```
> 
> Use
> 
> ```shell
> echo 123
> ```
> 
> ```
> ```shell
> echo 123
> ```
> ```

Hey @paramsiddharth is this more what you were thinking? https://github.com/ProgramEquity/amplify/wiki/Local-Setup

paramsiddharth:
> Hey @paramsiddharth is this more what you were thinking? [ProgramEquity/amplify/wiki/Local-Setup](https://github.com/ProgramEquity/amplify/wiki/Local-Setup)

Hi @smgraywood ! I meant, I was asking you to use block-level codeblocks than inline codeblocks. :)

smgraywood:
[...]

PR #479, opened 2023-02-03
Update README.md

beverand:
- Provided more context to the text around script/bootstrap. It was not clear what to do with that text.
- Added command to create the .env file.


paramsiddharth:
#466 can be discarded in favour of tis PR.

paramsiddharth (approved):
What would make it even better is documentation on the environment variables that need to be manually modified and set in the environment, but that's being covered under #467 and #431.

paramsiddharth:
Good job!

Issue #407, opened 2022-11-29
Donation Parsing / Custom Donation Input


DietBepis1:
Currently, Amplify only allows donations of $1, $2, $20, and $50 (and soon $1.50). It would be nice to allow users to make a custom donation, as well as make it more clear what the default donation is.

To accomplish this:
- [ ]  Update UI to accept custom donations and clearly indicate a default value
- [ ]  Update the '/create-checkout-session' route to accept a generic value.
- [ ]  Make sure that donations are correctly being sent to stripe, for example: $1 would be 100 because Stripe uses penny increments.
- [ ] Write tests for the above.

yoyoyojoe:
Although not entirely helpful as @Alex-is-Gonzalez mentioned during our meeting, just for the record here's the documentation on the Cicero API for its [`official`](https://cicero.azavea.com/docs/#officials) endpoint, and the expected response structure would be:

```js
{
  "response": {
    "errors": [],
    "messages": [],
    "results": {
      "count": {
        "to": number,
        "total": number,
        "from": number
      },
      "officials": [
        ... official objects ...
      ]
    }
  }
}
```

It'd certainly be nice if we can work with the actual response from the GET request, but for the time being I was able to move past the `404` & `AxiosError` on the `/campaign/` page (aka `SearchReps.vue` component by creating a `test_response` based on what properties were required from the existing code within `representatives.js`. And I was able to display the `DonateMoney.vue` component as how it currently looks:

<img width="330" alt="image" src="https://user-images.githubusercontent.com/86270711/216184514-65c29921-f69a-482e-8906-63dd5a2606af.png">

I'm not excited about pushing up a revision with mainly helpful notes and a `test_response` obj, but you can comment out the lines 114 to 130, add spaces as you'd like, and use the following code for bypassing the current error. 

```js
// starting line 129 after commenting out lines 114-130

    /* structure of data expected based on current code as of 02/01/2023:
    const response_results = {
      candidates: [
        {
          officials: [
            // rep: {}
            {
              office: {
                district: {
                  ocd_id: : 'string' || number
                },
                title: ''
              },
              chamber: {
                name_formal: 'string',
                name: 'string'
              },
              addresses: [
                // mainAddr: {}
                {
                  address_1: 'string',
                  address_2: 'string',
                  city: 'string',
                  postal_code: 'string',
                  state: 'string'
                }
              ],
              web_form_url: 'string',
              email_addresses: 'string',
              photo_origin_url: 'string',
[...]

Issue #430, opened 2022-12-18
Make `npm` scripts cross-platform


paramsiddharth:
# Make `npm` scripts cross-platform
Several scripts that use runtime environment variable initialization (`VAR=value cmd arg1 arg2`) would not work on Windows, and we've learnt that a number of our contributors are not necessarily Macintosh and Linux users but Windows users. Hence, this update is supposed to ease running the `npm` scripts for them.

## Scripts Affected
- `"dev:backend": "PORT=5000 nodemon -w ./server server/server.js"`
- `"lint": "DEBUG=eslint:cli-engine eslint --fix ."`
- `"lint:check": "DEBUG=eslint:cli-engine eslint ."`

## Note
[`cross-env`](https://www.npmjs.com/package/cross-env) is worth checking out. I've used it before for this particular cross-platform fix and it is pretty convenient and efficient.

dmilburn:
curious - why is this necessary inside of a codespace? do you also want to support outside of a codespace?

dmilburn:
we're currently working on making the codespaces experience more seamless and with a higher core count which should hopefully make it so this issue doesn't feel needed so we're going to punt on this one for now.

paramsiddharth:
> curious - why is this necessary inside of a codespace? do you also want to support outside of a codespace?

Hey Debbie! Apparently so, that was part of the discussion in the previous few cohorts. Codespaces aren't completely reliable and much of the big stories take a local development setup to develop, so it is only on us to make it easier to set up fresh development environments locally as well.

PR #498, opened 2023-02-15
Fix Part 1 for Issue #407

yoyoyojoe:
# Issue

**#407 Donation Parsing / Custom Donation Input:** 

Currently, Amplify only allows donations of $1, $2, $20, and $50 (and soon $1.50). It would be nice to allow users to make a custom donation, as well as make it more clear what the default donation is.

# Fix 

This PR primarily tackles step 1 / part 1 of 4 subtasks, which is to (1) update UI to accept custom donations and clearly indicate a default value, but paves the way for (3) that donations are correctly being sent to Stripe.

Here are some changes, assumptions, and consideration: 
- [x] added a "Custom Amount" `v-btn`
- [x] added a "Donation Amount" `v-text-input`
- [x] added corresponding input validation
  - [x] provides feedback message to user for 
    - No amount
    - Minimum amount ($1.50)
    - Maximum amount ($10,000)
  - [x] validated user input will always be a number with 2 decimal places
  - [x] if user tries to manually enter a number with 2+ decimal places, it'll be rounded up
  - [ ] todo: backend will only need to multiply by 100 for Stripe Checkout
- [x] applied consistent styles for all toggle buttons & text input
- [x] default value is $2, and user can input $1.5 as minimum value. So, we might be able to say the $1.50 button is solved together with this custom amount feature?
- [x] considered mobile responsiveness for different screen sizes
- [x] improved code readability

Enjoyed working with @coderlore & @Alex-is-Gonzalez on this!

Feedback is always welcome!

### DIFF
--- a/server/routes/api/representatives.js
+++ b/server/routes/api/representatives.js
@@ -45,6 +45,7 @@ const JURISDICTION_FILTER_MAP = {
   local: ['LOCAL_EXEC', 'LOCAL'],
   school: ['SCHOOL']
 }
+
 const ALLOWED_JURISDICTION_FILTERS = Object.keys(JURISDICTION_FILTER_MAP)
 
 // Endpoints
@@ -76,7 +77,7 @@ router.get('/:searchText', async (req, res) => {
   if (!isValidZIPcode && !isValidAddress) {
     res.status(400).send({
       error:
-        'Invalid zip code or street address format, valid examples of a ZIP code are 84054-6013 or 84054. The zipcode/street address used was ' +
+        'Invalid zip code or street address format, valid examples of a ZIP code are 84054-6013 or 84054. The zip-code/street address used was ' +
         searchText
     })
     return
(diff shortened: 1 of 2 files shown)


yoyoyojoe:
> 

Sure! I thought it was an entirely personal preference to send, for example, either `1.5` or `150` for a $1.5 donation to the backend. I am going with the preference as it currently is in `/server/routes/api/checkout.js`. I guess it is technically done, as the backend currently multiples the received value by 100 so that a $1.5 donation would be passed to Stripe as `150` in value, but I just thought since that file still needs to be worked on and improved, I would leave that one unchecked.

https://github.com/ProgramEquity/amplify/blob/57701dec925903c16930537098e9249e49269301/server/routes/api/checkout.js#L52-L56

The `checkout.js` file will be the next thing to work on as **_Fix Part 2 for Issue #407_**. I figured I should to submit this PR first before making more changes in other files and avoid having a PR with large differences in multiple files. That's why. Thanks @paramsiddharth!

paramsiddharth:
> Sure! I thought it was an entirely personal preference to send, for example, either `1.5` or `150` for a $1.5 donation to the backend. I am going with the preference as it currently is in `/server/routes/api/checkout.js`. I guess it is technically done, as the backend currently multiples the received value by 100 so that a $1.5 donation would be passed to Stripe as `150` in value, but I just thought since that file still needs to be worked on and improved, I would leave that one unchecked.

I understand. :) For the improvements you'd want in the `checkout.js` file, feel free to create new issues and work on them.

> The `checkout.js` file will be the next thing to work on as **_Fix Part 2 for Issue #407_**. I figured I should to submit this PR first before making more changes in other files and avoid having a PR with large differences in multiple files. That's why. Thanks @paramsiddharth!

That makes sense! :D Thank you for helping keep things organized — I do admit it is challenging to deal with conflict-resolution in PRs with huge diffs. Keeping PRs somewhat granular and focused on one issue at a time definitely helps. ✨

I'll test out the PR locally and give a review soon. Keep up the good work!

paramsiddharth:
Hey @yoyoyojoe! How is the PR going so far? :)

server/routes/api/representatives.js
```diff
@@ -31,7 +31,7 @@ const STORAGE = buildStorage({
     CACHE.delete(key)
```
DietBepis1, line 31: Changing/cleaning up the representatives route is not related to the donation functionality and should be a different PR. It is best practice to keep PRs very small and focused.

server/routes/api/representatives.js
```diff
@@ -89,114 +90,169 @@ router.get('/:searchText', async (req, res) => {
...
       cache: {
         // TODO: we are disabling the cache to test results first
         ttl: 1000 * 60 * 60 * 24 * 7 // the time until the cached value is expired in milliseconds: set to 1 week
       }
-    })
-
-    // if you want to check if response was cached, uncomment: this is a way to track the issue
-    // console.log('isCached:', cached)
-
-    const { errors, results } = response
```
DietBepis1, line 132: You may need to rethink what is happening here. In my local testing, response is undefined and attempting to search crashes the server.

DietBepis1 (changes requested):

server/routes/api/representatives.js
yoyoyojoe: Thanks for the feedback. I was secretly hoping changes in `representaties.js` could be lumped together with this PR without creating another issue, but I agree I should've kept the changes in `representatives.js` separate. I'll definitely try the "draft PR" next time to elicit preliminary feedback! Shall I go ahead and revert those commits?

server/routes/api/representatives.js
yoyoyojoe: This is the code that was deleted / modified. And you are correct, and this is what I've experienced repeatedly while working on `DonateMoney.vue` when the `CICERO_API_KEY` does not work. This is likely what prompted me to refactor `representatives.js` and lump the changes to this PR even when this is technically not within the same issue. 

Did you add `NODE_ENV='development'` in your `.env` file? (I have mine as `.env.dev`.) I've added this particular way to handle the error, but am definitely open to feedback. This is specifically to avoid crashing the server when developing. We can further discuss what should happen if the same request fails in production (e.g., what do we want the client to see? Do we want to redirect to a different page? Or do we want to render a message based on the error response sent back from the server?)... But I did stop myself there. 🐢 

I'll point to my changes on lines 131-179 here, and make some comments there:

https://github.com/ProgramEquity/amplify/blob/1ea6640c76fca1390c19c0672c299b74faa40058/server/routes/api/representatives.js#L131-L179

server/routes/api/representatives.js
```diff
@@ -89,114 +90,169 @@ router.get('/:searchText', async (req, res) => {
...
-          socialMediaPages: getOfficialSocialMediaPages(rep.identifiers),
-          photoCroppingCSS: getPhotoCroppingValues(rep.photo_cropping)
+  let response
+
+  if (error) {
+    // http://127.0.0.1:8080/api/representatives/91765 get request failed
+    // uncomment only when debugging
+    // console.error(error)
+
+    // to avoid server crashing so front-end work can continue, add NODE_ENV='development' to your .env
```
yoyoyojoe, line 136: This is the only hint to add `NODE_ENV='development'` to your .env file.

server/routes/api/representatives.js
```diff
@@ -89,114 +90,169 @@ router.get('/:searchText', async (req, res) => {
...
+    if (process.env.NODE_ENV === 'development') {
+      // a test_response helpful for front-end work when axios requests are not working
+      const test_response = {
```
yoyoyojoe, line 139: This is the variable that allows frontend to move to the next step even when getting the representatives request failed. I figured we only want to do this when we are in development mode.

server/routes/api/representatives.js
```diff
@@ -89,114 +90,169 @@ router.get('/:searchText', async (req, res) => {
...
+              ]
+            }
+          ]
         }
+      }
 
-        return repInfo
+      response = test_response
+    } else {
+      res.status(500).send({
```
yoyoyojoe, line 181: Now looking at this, I think line 181 warrants a `return` before the `res.status(500).send`. This is probably why I was seeing `Error: Can't set headers after they are sent to the client` at some point.

server/routes/api/representatives.js
```diff
@@ -89,114 +90,169 @@ router.get('/:searchText', async (req, res) => {
...
+  }
 
-    res.send(representatives)
-  } catch (error) {
-    console.log(error)
-    res.status(500).send({ error: 'Whoops' })
+  // if you want to check if response was cached, uncomment: this is a way to track the issue
+  // console.log('isCached:', cached)
+
+  const { errors, results } = response
```
yoyoyojoe, line 192: This should be assigned to be `test_response` if `NODE_ENV='development'` was added to `.env`.

yoyoyojoe:
> Hey @yoyoyojoe! How is the PR going so far? :)

@paramsiddharth Hi 😄  I was awaiting your feedback, but in the meantime I made some additional changes (sorry 🙄). Specifically for `DonateMoney.vue` I think it looks good and should work well and I am hoping to hear that you'd agree. I just learned about `inputmode="numeric"` and thought I should add it before you approve my last set of commits.

Other than that, I made some changes to `representatives.js`, primarily due to the frustration of not having my `CICERO_API_KEY` working. As Glenn mentioned, it is a separate issue. Regardless, I'm willing to make improvements based on feedback or revert those commits and wait to address those in a separate issue.

server/routes/api/representatives.js
DietBepis1: Yes, got ahead and revert this part. You can work on improving this as your next issue. If you've never used `git cherry-pick`, this could be a good opportunity to mess with it. Let me know if you need help.

https://git-scm.com/docs/git-cherry-pick

src/components/DonateMoney.vue
```diff
@@ -8,13 +8,71 @@
...
+          :min="1.5"
+          label="Donation Amount"
+          :value="customDonationAmount"
+          inputmode="numeric"
+          @input="validateDonationAmount"
+          required
+        />
+      </div>
+
+      <p v-if="message !== ''" class="message">
```
DietBepis1, line 73: You can do `v-if="message"` since an empty string is falsey in JavaScript.

src/components/DonateMoney.vue
```diff
@@ -25,22 +83,75 @@
...
-      return {
-        donation: 1.50
-      }
+  name: 'DonateMoney',
+  props: [],
+  data () {
+    return {
+      donationAmount: 2,
+      customAmountSelected: false,
+      customDonationAmount: undefined, // set min value to be 1.5
```
DietBepis1, line 92: If you're going to use this to set a minimum value, then don't set to undefined. Undefined will cause an exception if customDonationAmount is not set when it is first used.

server/routes/api/representatives.js
yoyoyojoe: I just tried it with Param's last update, but it doesn't seem to be working for me. I wonder why I'd be an exception though. Wouldn't everyone be having the same issue?

Despite having issues here, I'm still willing to revert these changes.

src/components/DonateMoney.vue
yoyoyojoe: > If you're going to use this to set a minimum value, then don't set to undefined. 

I could remove the comment if that's the source of confusion. It was a deliberate choice to use undefined because I think it makes sense from the user's perspective, that if one was to choose a custom amount, it should be left blank until the user fills it in. The minimum of $1.5 is also a requirement of the feature, as it's the least amount of donation acceptable to offset the cost of sending a letter.

> Undefined will cause an exception if customDonationAmount is not set when it is first used.

That's why `validateDonationAmount` is put in place to handle these exceptions, including when the user is trying to submit an empty value, a value less than $1.50, or a value more than $10,000.00 (I made that maximum limit up. Can be changed to anything that makes more sense.)

server/routes/api/representatives.js
paramsiddharth: @yoyoyojoe @DietBepis1 For the `NODE_ENV` part, we can plan setting that variable in the `dev` script under #430. I'd already known we'll come across this at some point, haha.

@yoyoyojoe You are right, there is an issue with the key and we're talking about it. Thank you for reporting, I'll make sure we get it fixed as soon as possible. Till then you can use your own or Glenn's trial key.

server/routes/api/representatives.js
paramsiddharth: @yoyoyojoe Good point! Do you think this should be related to #509? Maybe for `NODE_ENV === 'development'` we can choose to hit the mock endpoint instead of the actual API.

paramsiddharth:
Hi @yoyoyojoe ! Great work, I believe there's a lot of discussion we need to do and I believe we need to better utilize the pair hours to sort this out. If needed, I'll ask Manisha to allow us to have our own separate breakout room. :)

yoyoyojoe:
Will continue working on PR #523 523

PR #499, opened 2023-02-16
added previousPage in checkout route

masmei:
Changed the "cancel" routing on stripe checkout page to go back to "previous page" that the user was on. Still need to figure out how to repopulate the selections/data previously entered.

### DIFF
--- a/server/routes/api/checkout.js
+++ b/server/routes/api/checkout.js
@@ -61,6 +61,7 @@ router.post('/create-checkout-session', async (req, res) => {
     }
 
     const origin = req.get('origin')
+    const previousPage = req.header('referer')
 
     const session = await stripe.checkout.sessions.create({
       line_items: [
@@ -78,7 +79,7 @@ router.post('/create-checkout-session', async (req, res) => {
       mode: 'payment',
       allow_promotion_codes: true,
       success_url: origin + '/complete?session_id={CHECKOUT_SESSION_ID}',
-      cancel_url: origin
+      cancel_url: previousPage
     })
 
     res.json({ url: session.url, sessionId: session.id })


DietBepis1 (commented):
Try doing something like this to restore the localstorage data:

```
const session = await stripe.checkout.sessions.create({
...
  cancel_url: `${previousPage}?cancelled-session={CHECKOUT_SESSION_ID}`
})
```

You may need to slightly modify the path of referrer in this instance. For example if there is already a query parameter.

yoyoyojoe:
What does adding the `cancelled-session={CHECKOUT_SESSION_ID}` do?

Is it helpful to restore to the exact last state of the previousPage?

paramsiddharth:
> What does adding the `cancelled-session={CHECKOUT_SESSION_ID}` do?
> 
> Is it helpful to restore to the exact last state of the previousPage?

@yoyoyojoe Yes! If I'm right, it uses that session ID as a key to the `localStorage` object to retrieve the state/data. @DietBepis1 Could you elaborate?

paramsiddharth (dismissed):

DietBepis1:
> > What does adding the `cancelled-session={CHECKOUT_SESSION_ID}` do?
> > Is it helpful to restore to the exact last state of the previousPage?
> 
> @yoyoyojoe Yes! If I'm right, it uses that session ID as a key to the `localStorage` object to retrieve the state/data. @DietBepis1 Could you elaborate?

@paramsiddharth is correct. We are storing the state of the application in `localStorage` as a key value pair, like in this screenshot:

localstorage

The storage happens with the Vuex actions in /src/store/index.js called `dumpStateToLocalStorage` and `retrieveStateFromLocalStorage`. Essentially those actions just use the native API of the browser to store and read a json string that we convert our state object into. Then it will delete that localStorage entry. Of course, the screenshot above shows ones that were not deleted because something failed while I was coding :)

Let me know if more explanation is needed.

DietBepis1:
I pushed a small change to this branch to illustrate what the reconstitution of the Vuex state would look like. I think this can be merged once we rebase it.

paramsiddharth:
@DietBepis1 Hi Glenn, would we like to merge the PR with a final review after resolving conflicts?

PR #501, opened 2023-02-16
removal of bad example key for Stripe secret.

smgraywood:
**Issue:
Appears that Stripe secret key was committed accidentally

**Fix:
removing out of abundance of caution

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,7 +23,7 @@ AUTH0_AUDIENCE=your_Auth0_identifier_value
 AUTH0_DOMAIN=your_Auth0_domain
 
 # Stripe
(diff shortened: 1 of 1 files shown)


smgraywood:
> Hi Sarah! :D The first time I saw it, I also thought it was accidentally committed, haha. 😁 But in this case it is a test key that helps centralize development configuration for the Stripe flow, and it is important that the test key be committed to the code in the repository, so it's alright and not a vulnerability. :)

Ah ok, makes sense. Figured I would just be proactive in case it was a security vulnerability :)

paramsiddharth:
> > Hi Sarah! :D The first time I saw it, I also thought it was accidentally committed, haha. 😁 But in this case it is a test key that helps centralize development configuration for the Stripe flow, and it is important that the test key be committed to the code in the repository, so it's alright and not a vulnerability. :)
> 
> Ah ok, makes sense. Figured I would just be proactive in case it was a security vulnerability :)

Thank you, Sarah! I'd actually encourage that. It's an open-source community and we never know when there might be a vulnerability all along that none of us noticed. :) Feel free to create new issues if you come across anything suspicious.

PR #506, opened 2023-02-22
update Contributing documentation and move into root

christina-ml:
Update currently used versions of Node.js and npm
Move CONTRIBUTING.md out of .github/ and into root


paramsiddharth (approved):
Looks good to me!

PR #508, opened 2023-02-22
Adding new documentation issue type

### LINKED ISSUE #505: Create issue template for documentation issues
The current issue templates are kind of overkill for documentation.  We should create a simple one that prompts the contributor to describe what needs documentation and applies the `Documentation` label.  We have an open issue to [convert our issue templates to forms](https://github.com/ProgramEquity/amplify/issues/355) so it would be great to use that syntax here.

knowaveragejoe:
Resolves #505


paramsiddharth:
We should also have a checklist that allows the issue creator to specify which category/categories the documentation issue might fall into!

therzka:
👋 @knowaveragejoe would you like to keep working on this or is it OK for someone else to take over?

PR #510, opened 2023-02-24
Modifications to the bootstrap script to properly run on OS X

### LINKED ISSUE #500: Create MacOS Script for Local Development
We need to create a script for MacOS to run using Homebrew to make starting local development simpler for MacOS users.

We can do this by making a new script based off of the script/bootstrap script currently used to set up development with codespaces.

knowaveragejoe:
Currently the logic in `setup/bootstrap.sh` causes the script to fail on OS X. This is because it currently is trying to use linuxbrew when it should not. This change adds some conditions to ensure the linuxbrew commands only run on Linux or in Github Codespaces.

I tested this against my fork using Codespaces and it seems to be working. 

Resolves #500

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,2 +1,2 @@
 brew "nodenv"
-brew "postgresql"
+brew "postgresql@14"
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -2,12 +2,13 @@
 
 set -e
 
+# If homebrew is not installed, install it
 command -v brew >/dev/null 2>&1 || {
   echo "==> Installing Homebrew..."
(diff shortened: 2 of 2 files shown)


paramsiddharth:
I believe fixing the version was important – I had to upgrade my PostgreSQL installation (migrating the existing databases is tough) when I started working on Amplify. Great job, @knowaveragejoe ! You can ignore my comment on #500 asking you to add a link to the file.

DietBepis1:
@paramsiddharth do you need someone else to test? I can try this on my Macbook.

paramsiddharth:
@DietBepis1 Sure! That would help a lot. Just make sure you go through the script first so that it doesn't conflict with your existing setup/environment.

PR #512, opened 2023-03-01
Validate User Form Before  Donation

JimmyAyivor:


src/components/SignName.vue
```diff
@@ -10,6 +10,7 @@
           label="Full Name"
           placeholder="John Doe"
           required
+          @keyup="checkFormFilled"
```
DietBepis1, line 13: It is an interesting and clever solution to check on every keystroke if the form is filled. I don't necessarily think this needs to be changed, but consider that every method call here incurs a performance penalty (even if it is a small one).

src/components/SignName.vue
```diff
@@ -182,7 +201,12 @@ export default {
             .catch(function (error) {
                 console.log(error)
             })
-        }
+        },
+        checkFormFilled() {
+      // check if the form is fully filled here
+      this.isFormFilled = !!this.checkForm()
+      bus.$emit('formFilled', this.isFormFilled);
```
DietBepis1, line 208: I don't think you need to import bus. You should have access to emit with `this.$emit`

src/components/SignName.vue
```diff
@@ -137,6 +146,16 @@ export default {
     },
 
     methods: {
+      checkForm() {
+      const requiredFields = document.querySelectorAll('input[required], textarea[required]');
```
DietBepis1, line 150: I think that this could be simplified by using the `form` computed property that already exists. For example
```
for ([formValue] in Object.values(this.form)) {
  if (!formValue) {
    // some logic
    }
}
```
This may not work exactly, but just a suggestion.

paramsiddharth:
Hey @JimmyAyivor ! :) Would you like to finish up on Glenn's review of the PR and the requested changes? We can successively merge your PR once that's done.

paramsiddharth (approved):
Looks good to me! :) Good work, Jimmy.

src/components/SignName.vue
DietBepis1, line 13: This works as intended, but I was thinking we could clean up some of the repeated code if we used a watcher on the form property. I tried a quick and dirty test and it works how I think it should. There is another watcher in `SignName.vue` that you could use as an example. The implementation would be essentially the same.

https://vuejs.org/guide/essentials/watchers.html

src/components/TakeAction.vue
```diff
@@ -78,17 +78,17 @@
         </v-expansion-panel-header>
 
         <v-expansion-panel-content>
-          <sign-name />
+          <sign-name @form-filled="handleFormFilled" />
```
DietBepis1, line 81: Awesome!

src/components/TakeAction.vue
```diff
@@ -78,17 +78,17 @@
...
         <v-expansion-panel-content>
           <v-btn
             width="160"
-            dark
             color="primary"
+            :disabled="!formFilled"
             @click="nextPage({ userData })"
           >
-            Next
+            {{ formFilled ? 'Next' : 'Please Fill Form!' }}
```
DietBepis1, line 91: I think we could just disable the button and not change the text. Perhaps along with adding a disabled cursor effect. 

https://www.w3schools.com/cssref/pr_class_cursor.php

DietBepis1 (changes requested):

src/components/TakeAction.vue
JimmyAyivor: Thank you, I will do just that.

src/components/TakeAction.vue
JimmyAyivor: Thank you!

src/components/SignName.vue
JimmyAyivor: Thank you for this suggestion, I will look into it and make the necessary updates.

src/components/TakeAction.vue
```diff
@@ -78,17 +78,16 @@
...
         <v-expansion-panel-content>
           <v-btn
             width="160"
-            dark
             color="primary"
+            :disabled="!formFilled"
             @click="nextPage({ userData })"
           >
-            Next
+            {{ formFilled ? 'Next' : 'Next' }}
```
DietBepis1, line 90: If the button text is always `Next`, is the ternary operator really necessary? 😃

DietBepis1 (changes requested):

src/components/TakeAction.vue
JimmyAyivor: lol 😃

src/components/SignName.vue
```diff
@@ -100,91 +100,108 @@ import axios from 'axios'
...
+        email: this.email
       }
+    }
+  },
+
+  watch: {
+    name() {
+      this.errorMessages = ''
     },
+    form: {
```
JimmyAyivor, line 138: Do tell me what you think, I would to know if  this is a right implimentation of a watcher in vue

PR #520, opened 2023-03-02
Created welcome Message yml

Alex-is-Gonzalez:
Created welcome message upon creating a first issue and first pr.


paramsiddharth (approved):
Looks good to me!

@JamesMGreene Do you think we'd be able to test this out using your second account, once merged?

.github/workflows/welcome_message.yml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+jobs:
+  welcome-new-contributor:
+    runs-on: ubuntu-latest
+    steps:
```
JamesMGreene, line 12: I recommend adding [a `timeout-minutes` configuration](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) for **every** Actions job to ensure you don't end up wasting the default 6 hours in the case of a runaway Action/job:

```suggestion
    timeout-minutes: 2
    steps:
```

.github/workflows/welcome_message.yml
```diff
@@ -0,0 +1,18 @@
+name: 'Welcome New Contributors'
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
```
JamesMGreene, line 8: It's a best practice to restrict the permissions granted to the `GITHUB_TOKEN` to just the minimum set that are relevant to the workflow/job:

```suggestion

permissions:
  issues: write
  pull-requests: write

```

.github/workflows/welcome_message.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  welcome-new-contributor:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Greet the contributor'
+        uses: garg3133/welcome-new-contributors@v1.2
```
JamesMGreene, line 14: From a security perspective, I would recommend you pin to a specific commit of the Action to ensure its tag (e.g. `v1.2`) is not maliciously swapped out from under you without giving you the opportunity to review it:

```suggestion
        uses: garg3133/welcome-new-contributors@a38583ed8282e23d63d7bf919ca2d9fb95300ca6
```

JamesMGreene (commented):
This workflow looks good overall, though I left a few suggestions for best practices.

I don't see anything here that would warrant the need to run it in a separate account first, so just make sure to keep an eye out for potential failing runs in the future. 👍🏻

paramsiddharth (changes requested):
@Alex-is-Gonzalez I think it's best we move forward with James's advice and update the workflow. :)

Alex-is-Gonzalez:
@JamesMGreene Thank you for the thorough feedback! I learned a lot from this, I went ahead and changed the yml based on the notes provided!

JamesMGreene (approved):
LGTM. Thanks for following up on my suggestions! ❤️

Issue #449, opened 2023-01-07
Updating the Campaign Model


DietBepis1:
To accommodate the addition of Single Campaign Mode, we need to add some fields to Campaign. The new schema will be something like:

```
{
  "campaign": {
    "id": 1,
    "organization": "Sogorea Te' Land Trust",
    "name": "Sogorea Te' Land Trust",
    "cause": "Civic Rights",
    "type": "Grant",
    "page_url": "[www.thebreatheact.org](https://sogoreate-landtrust.org/)",
    "campaign_tagline": "Protect Our Sacred Shellmounds",
    "campaign_text": "Lorem ipsum dolor sit amet.....",
    "supplemental_text": "Some other text blurb that appears on the search reps page"
  }
```

This will simplify our configuration when deploying Amplify for a single campaign.

paramsiddharth:
@DietBepis1 Hi Glenn, I have a question. Will the updated schema apply to all the campaigns in the campaigns table, or just to a single campaign in a separate table?

DietBepis1:
I was planning to add nullable fields to the current model, but would be open to discussing other options if you have an idea.

smgraywood:
Hi @DietBepis1 ! I updated the schema to reflect the requested changes. Are we getting rid of the logo_url, letters_goal and donation_goal fields? Also does this need to be linked up with anything in the frontend or are we just updating this schema for this ticket?

paramsiddharth:
@smgraywood Awaiting confirmation from @DietBepis1, but I believe directly editing the migration is not how we should do this. We need to create a new migration using the [stub](https://github.com/ProgramEquity/amplify/blob/main/server/db/_migration.stub.js) and update it to add/remove and modify columns and/or types as needed.

Issue #406, opened 2022-11-29
Implement logging


DietBepis1:
We are at the point where it is becoming imperative to be able to monitor Amplify's production deployments. During the Sogorea Te' onboarding event, we received dozens of donations, but many of them ultimately failed to process and refunds were issued. Those refunds are potentially hundreds or thousands of donation dollars that may be lost forever. If we had logging, those issues may have been able to be resolved in the moment.

Proposal: Implement logging using [Winston](https://github.com/winstonjs/winston). It is a non-blocking logging library that can log to anything from a plain text file to SASS services like [Datadog](https://www.datadoghq.com/).

To Do:
- [ ]  Set up base Winston.
- [ ] Implement and document logging levels + Amplify best practices.
- [ ] Research logging services and what logging ability that Heroku can give.
- [ ] Integrate with chosen provider.

rsensenig:
I'm currently working on reading up on the Winston documentation. My next step will be then to implement Winston on a simple Vue app that I created to test out how it works, and then work on setting up base Winston for Amplify.

PR #522, opened 2023-03-07
Updating the Campaign Model

smgraywood:
Responds to [Updating the Campaign Model #449](https://github.com/ProgramEquity/amplify/issues/449)

##Issue

Schema and model needed to be updated to accommodate the addition of Single Campaign Mode
 
##Fix

Add fields to campaign.js and 20210513215151_create-campaigns-table.js. Additionally make some fields nullable to accommodate the new schema.


server/db/migrations/20210513215151_create-campaigns-table.js
```diff
@@ -8,11 +8,15 @@ module.exports = {
...
+      table.string('name').notNullable()
+      table.string('cause').notNullable()
+      table.string('type').notNullabel()
       table.text('page_url').notNullable()
-      table.integer('letters_sent').notNullable()
-
+      table.string('campaign_tagline').notNullable()
+      table.string('campaign_text').notNullable()
+      table.string('supplemental_text').nullable()
+      table.integer('letters_sent').nullable()
```
paramsiddharth, line 19: @DietBepis1 Do you think we should modify an existing migration for this? Shouldn't we create a new migration?

server/db/migrations/20230309195953_update-campaign-model.js
```diff
@@ -0,0 +1,20 @@
+const tableName = 'campaigns'
+
+module.exports = {
+  async up(knex) {
+    await knex.schema.createTable(tableName, (table) => {
```
DietBepis1, line 5: You are on the right track, but you will need to modify the table and not create a new one with the same name. Look at the [alterTable](https://knexjs.org/guide/schema-builder.html#altertable) method.

DietBepis1 (changes requested):

server/db/migrations/20230309195953_update-campaign-model.js
smgraywood: Thank you! Ok, I switched the function over from createTable to alterTable.

server/db/migrations/20230309195953_update-campaign-model.js
```diff
@@ -0,0 +1,20 @@
+const tableName = 'campaigns'
+
+module.exports = {
+  async up(knex) {
+    await knex.schema.alterTable(tableName, (table) => {
+      // Auto-incrementing non-nullable unsigned integer primary key "id" field
+      table.increments()
```
DietBepis1, line 7: If you are altering the table, there is no need to add `increments()` again. It will cause a duplicate key error.

server/db/migrations/20230309195953_update-campaign-model.js
```diff
@@ -0,0 +1,20 @@
...
+
+      // Simple fields
+      table.string('campaign_tagline').notNullable()
```
DietBepis1, line 10: I think these can be nullable, because we currently only need them for single-campaign mode.

server/db/migrations/20230309195953_update-campaign-model.js
```diff
@@ -0,0 +1,17 @@
...
+      // Simple fields
+      table.string('campaign_tagline').nullable()
+      table.string('campaign_text').nullable()
+      table.string('supplemental_text').nullable()
+    })
+  },
+
+  async down(knex) {
+    // Drop the table
+    await knex.schema.dropTable(tableName)
```
paramsiddharth, line 15: I don't think it should drop the table here, Sarah.

Tip: `down(knex)` is supposed to undo the updates made by `up(knex)`. 😉

paramsiddharth (changes requested):

rsensenig:
Hi @paramsiddharth - I decided to pick up this issue since I'm familiar with knex migrations and objection models from working on [Implement logging#406](https://github.com/ProgramEquity/amplify/issues/406). Other than adding back the down(knex) function, are there any other changes to be made before this PR is merged? Thanks!

paramsiddharth:
> Hi @paramsiddharth - I decided to pick up this issue since I'm familiar with knex migrations and objection models from working on [Implement logging#406](https://github.com/ProgramEquity/amplify/issues/406). I have added back the down(knex) function like you requested. Are there any other changes to be made before this PR is merged? Thanks!

Hi @rsensenig — Good to see you picking up this issue! I think you need to look carefully at the `down` function. The `up` function alters the table in a certain manner. `down` should undo it i.e. If it is creating certain columns, it should delete them. 😊 Think carefully.

rsensenig:
> > Hi @paramsiddharth - I decided to pick up this issue since I'm familiar with knex migrations and objection models from working on [Implement logging#406](https://github.com/ProgramEquity/amplify/issues/406). I have added back the down(knex) function like you requested. Are there any other changes to be made before this PR is merged? Thanks!
> 
> Hi @rsensenig — Good to see you picking up this issue! I think you need to look carefully at the `down` function. The `up` function alters the table in a certain manner. `down` should undo it i.e. If it is creating certain columns, it should delete them. 😊 Think carefully.

Thanks Param! I looked at the knex 'down' function again and updated it so that it drops the columns. Please let me know if this is correct now.

server/db/models/campaign.js
```diff
@@ -23,13 +23,9 @@ class Campaign extends BaseModel {
...
         type: { type: 'string', enum: ['Starter', 'Accelerator', 'Grant'] },
         page_url: { type: 'string', minLength: 1 },
-        logo_url: { type: 'string', minLength: 1 },
-        letters_goal: {
-          anyOf: [{ type: 'integer', minimum: 0 }, { type: 'null' }]
-        },
-        donation_goal: {
-          anyOf: [{ type: 'integer', minimum: 0 }, { type: 'null' }]
-        }
+        campaign_tagline: { type: 'string', minLength: 1 },
```
DietBepis1, line 26: I think that this will need a `minLength` of 0, or to omit that property if we want to have an empty string.

DietBepis1 (changes requested):

server/db/models/campaign.js
rsensenig: @DietBepis1 I've updated the minLength to 0. I did it this way so that when future contributors look at the code they'll see we purposefully set it to that, so that the tagline is optional.

paramsiddharth:
@DietBepis1 Hey Glenn! What do you think is the best way to make the property optional? I don't believe `minLength: 0` should be the way to go  — Let me know what you think!

References:
- https://stackoverflow.com/questions/16204871/what-is-the-difference-between-required-vs-optional-in-json-schema
- https://vincit.github.io/objection.js/api/model/static-properties.html#static-jsonschema

rsensenig:
> @DietBepis1 Hey Glenn! What do you think is the best way to make the property optional? I don't believe `minLength: 0` should be the way to go — Let me know what you think!
> 
> References:
> 
> * https://stackoverflow.com/questions/16204871/what-is-the-difference-between-required-vs-optional-in-json-schema
> * https://vincit.github.io/objection.js/api/model/static-properties.html#static-jsonschema

After discussing and researching with @paramsiddharth I agree that minLength should be removed instead of being set to 0, so I went ahead and made that change. Let us know whether this looks good to you @DietBepis1.

server/db/models/campaign.js
```diff
@@ -23,13 +23,9 @@ class Campaign extends BaseModel {
...
         page_url: { type: 'string', minLength: 1 },
-        logo_url: { type: 'string', minLength: 1 },
-        letters_goal: {
-          anyOf: [{ type: 'integer', minimum: 0 }, { type: 'null' }]
-        },
-        donation_goal: {
-          anyOf: [{ type: 'integer', minimum: 0 }, { type: 'null' }]
-        }
+        campaign_tagline: { type: 'string' },
+        campaign_text: { type: 'string', minLength: 1 },
```
DietBepis1, line 27: I think omitting the `minLength` like @paramsiddharth said is the right move. Let's do it on `campaign_text` also and this should be ready to merge.

DietBepis1 (changes requested):

server/db/models/campaign.js
beverand: Sorry for the delayed reply on this thread. If it's optional you can handle that by setting a default value or making it a nullable field.

DietBepis1 (approved):

PR #523, opened 2023-03-08
407 custom donation input 2

yoyoyojoe:
# Issue

**#407 Donation Parsing / Custom Donation Input:** 

Version 2 of PR #498 

Currently, Amplify only allows donations of $1, $2, $20, and $50 (and soon $1.50). It would be nice to allow users to make a custom donation, as well as make it more clear what the default donation is.

# Fix 

This PR primarily tackles step 1 / part 1 of 4 subtasks, which is to (1) update UI to accept custom donations and clearly indicate a default value, but paves the way for (3) that donations are correctly being sent to Stripe.

Here are some changes, assumptions, and consideration: 
- [x] added a "Custom Amount" `v-btn`
- [x] added a "Donation Amount" `v-text-input`
- [x] added corresponding input validation
  - [x] provides feedback message to user for 
    - No amount
    - Minimum amount ($1.50)
    - Maximum amount ($10,000)
  - [x] validated user input will always be a number with 2 decimal places
  - [x] if user tries to manually enter a number with 2+ decimal places, it'll be rounded up
  - [x] todo: backend will only need to multiply by 100 for Stripe Checkout
- [x] applied consistent styles for all toggle buttons & text input
- [x] default value is $2, and user can input $1.5 as minimum value. So, we might be able to say the $1.50 button is solved together with this custom amount feature?
- [x] considered mobile responsiveness for different screen sizes
- [x] improved code readability

Enjoyed working with @coderlore & @Alex-is-Gonzalez on this!

Feedback is always welcome!


src/components/DonateMoney.vue
```diff
@@ -25,22 +83,75 @@
...
+        this.message = ""; // only refreshes message after input value becomes valid
+      } else {
+        this.message = "The donation amount is not valid."; // edge case?
+        return;
+      }
     },
-    methods: {
-      submit () {
-        axios.post('/api/checkout/create-checkout-session', {donationAmount: this.donation})
+    createCheckoutSession (donationAmount) {
```
DietBepis1, line 152: Very nice. The whole front-end appears to be working correctly. Great job handling the cases. There appears to be some work left in the `create-checkout-session` call -- namely, a check for valid donation amount that must be changed.

src/components/DonateMoney.vue
```diff
@@ -25,22 +83,75 @@
...
+      }
+      if (value < 1.5) {
+        this.message = "Please enter a donation amount higher than $1.50.";
+      }
+      if (value > 10000) {
+        this.message = "Amplify currently only accept donation amounts less than $10,000."
+      }
+
+      isValid = false;
+      return { value, isValid }; // i.e. { undefined, false }
```
DietBepis1, line 138: This works, but seems unnecessarily complicated. The method name implies that the function is just doing validation, but there is some other logic happening which is confusing. Consider splitting this into a pure validation method and a method to handle whether or not the custom donation is being used.

src/components/DonateMoney.vue
```diff
@@ -8,13 +8,71 @@
...
+          elevation="2"
+          raised
+          :value="20"
+          @click="unsetCustomAmountSelection"
+        >
+          20
+        </v-btn>
 
-        <v-btn elevation="2" :value="50"> 50 </v-btn>
+        <v-btn
```
DietBepis1, line 36: Looks good!

DietBepis1 (changes requested):

src/components/DonateMoney.vue
yoyoyojoe: I was going to work on that in a different PR. Would you like me to make that change here?

src/components/DonateMoney.vue
yoyoyojoe: Hi Glenn, thanks for the feedback. 

In my mind, what the validate function should return is a boolean and the acceptable value. I figured the user feedback is important as well. As I'm trying to understand which part is unnecessarily complicated, and I'm guessing perhaps you are referring to how I am additionally checking if custom amount is selected inside of `validateDonationAmount`, so lines 116-120:

https://github.com/ProgramEquity/amplify/blob/c0b1cb989b29e834c8a98b6a098d325128f930c1/src/components/DonateMoney.vue#L116-L120

A few things about this: 
(1) Do you think it's necessary to validate a value that is preselected? One approach can be that I can remove lines 116-120 because, arguably, a pre-selected value should not need to be validated. I don't know if there is (enough of) a security issue here. Removing those lines alone, Does this make the function a pure validation method?
(2) Or, perhaps you want me to separate the logic to provide user feedback from the validate function? Or did you want me to return only a boolean for this validate function? If the function returns a boolean, then somewhere else the number will need to be sanitized though.
(3) I checked out the `SignName` component and learned a few things about using `rules` and `ref`. As it currently is in `SignName`, I don't think it works though. `rules` do not validate nor does it sanitize the input though, although it provides a consistent visual for providing visual feedback to the user. I can try that implementation of the same feature. Would you like me to do that instead?

I can think of a few different ways to change my approach, but would love to get clarity on scope and expectations. Thanks!

src/components/DonateMoney.vue
yoyoyojoe: Made some changes in `checkout.js` and refactored for readability. Hope you like it!

paramsiddharth:
@yoyoyojoe Hey Joey! How is the PR going? I'd also like to remind you to update the branch with the changes from `main` so that it remains consistent and we minimize conflicts and resolve them as soon as we get them. :)

server/routes/api/checkout.js
```diff
@@ -41,27 +41,54 @@ router.post('/create-transaction', async (req, res) => {
...
+  
+  if (isNaN(value)) message = "Please select or enter a valid amount.";
+  if (value < 1.5) message = "Please enter a donation amount higher than $1.50.";
+  if (value > 10000) message = "Amplify currently only accept donation amounts less than $10,000."
+
+  inputIsValid = false;
+  return [inputIsValid, message]; // i.e. [ false, "Please select or enter a valid amount." ]
+}
+
+async function checkout(donationAmount, origin) {
```
DietBepis1, line 90: I don't think you need to separate the checkout logic from the route method. In this case, the route method's purpose is to create the checkout.

server/routes/api/checkout.js
```diff
@@ -41,27 +41,54 @@ router.post('/create-transaction', async (req, res) => {
...
 
-    const origin = req.get('origin')
+    return res.json({ url: session.url, sessionId: session.id })
+  } else {
+    return res.status(400).send({ error: 'Invalid Amount; did not create Stripe checkout session', message })
+  }
+})
+
+// format input value
+function format(value) {
```
DietBepis1, line 68: Good idea to abstract these validation and formatting methods from the route. Since you're also using them in the Vue component, you could create a directory for utility code and import the methods from there. That will dry this up a little bit.

DietBepis1 (changes requested):

server/routes/api/checkout.js
yoyoyojoe: Noted. I can incorporate this back to be directly within the post req without breaking it out into a function.

src/components/DonateMoney.vue
yoyoyojoe: I've created a `format` function, a `validate` function and saved them in `util` folder so that both frontend and backend can use them. In addition, although I liked my custom solution for providing user feedback on custom donation amount input, I'm going to use the `rules` property to invoke custom functions based on `v-text-field`'s state for some visual consistency. (ref: [validation & rules](https://vuetifyjs.com/en/components/text-fields/#validation-26-rules)) 

https://github.com/ProgramEquity/amplify/blob/ac3c112098522642d87d6c0167ef48285a3c0aff/src/components/DonateMoney.vue#L93

https://github.com/ProgramEquity/amplify/blob/ac3c112098522642d87d6c0167ef48285a3c0aff/src/components/DonateMoney.vue#L114

yoyoyojoe:
Noting that with the failing test, I have some questions/observations:

(1) In the logs, the failing Integration Tests is recommending to update `caniuse-lite`. Should I go ahead and do that?

> Browserslist: caniuse-lite is outdated. Please run:
>   npx browserslist@latest --update-db
>   Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
> 

(2) When I run the following test locally, it seems to run OK. Should I change something there?

```bash
npm test server/__tests__/integration/axios-cache-interceptor.test.js
```
image

(3) Is this test getting into the territory of testing `axios-cache-interceptor`?

yoyoyojoe:
The tests seemed to have passed after clicking on **Update branch**. 👍🏻 

> **This branch is out-of-date with the base branch** image. 
> Merge the latest changes from main into this branch. 

---

> Noting that with the failing test...

paramsiddharth:
> Noting that with the failing test, I have some questions/observations:
> 
> (1) In the logs, the failing Integration Tests is recommending to update `caniuse-lite`. Should I go ahead and do that?

It's a very common message I've noticed in many open-source projects! @DietBepis1 Do you think we should do it?

> The tests seemed to have passed after clicking on **Update branch**. 👍🏻
> 
> > **This branch is out-of-date with the base branch** image.
> > Merge the latest changes from main into this branch.
> 
> > Noting that with the failing test...

Looks like you got it to pass now. :)

paramsiddharth (dismissed):
Looks good to me! Good job! :)

P.S. Don't forget to rename the functions! 😉

server/routes/api/checkout.js
```diff
@@ -42,48 +42,56 @@ router.post('/create-transaction', async (req, res) => {
...
-  try {
-    const acceptableCharges = [1, 2, 20, 50]
-    const { donationAmount } = req.body || {}
-    const parsedDonationAmount = parseInt(donationAmount, 10)
+  const { donationAmount } = req.body || {}
+  const origin = req.get('origin')
 
-    let donation
+  const input = formatDonationAmount(donationAmount)
+  const inputIsValid = validateDonationAmount(input)
```
DietBepis1, line 49: Nice!

server/routes/api/checkout.js
```diff
@@ -42,48 +42,56 @@ router.post('/create-transaction', async (req, res) => {
...
+  const donationAmountForStripe = input * 100 // Stripe accepts values in cents
```
DietBepis1, line 50: You could move this into the try block and save a calculation. Not a big deal, however.

DietBepis1 (dismissed):

server/routes/api/checkout.js
yoyoyojoe: Agreed. It's an easy fix though.

yoyoyojoe: Actually, I'm thinking that it makes sense to move the computation inside of the `if` statement, not the `try` block. How would moving this into the `try` block save a calculation?

Or did you mean that creating a const donationAmountForStripe is not necessary? Again, readability was the only reason, and it'd also be a super easy fix.

DietBepis1: You are correct to move it inside the `if` block. I was thinking that it was a little tighter grouping inside the `try`, but now that I see what you did, I like that better.

I believe that `donationAmountForStripe` is necessary because we have to send the donation value in cents to the Stripe API.

DietBepis1 (approved):

paramsiddharth (approved):
Absolutely amazing work, Joey! <3

PR #524, opened 2023-03-11
Adds auth0-vue library.

wallace:
The solution here was to remove the pinned version for vue-router from package.json, add the new library, and let the npm install find a version of the library that worked for the existing libs and the auth0-vue library.

Once that was done, I re-added the pinned vue-router with `npm install --save vue-router` to pass the build check.


paramsiddharth:
@DietBepis1 Do you think this is a good long-term solution?

PR #529, opened 2023-03-16
Copy of pr428

SAUMILDHANKAR:
For this PR, I created a branch named ```copyOfPr428``` based of branch ```sammychinedu2ky/amplify:main``` and then merged the latest commits into branch ```copyOfPr428```.

The objective is to see if this PR will pass the integration tests which were failing earlier in PR #428


SAUMILDHANKAR:
@paramsiddharth Hi Param, just wanted to share with you that the integration test is still failing for this PR with similar error as in PR 428. The error mentioned to use Test LOB API key, so just wanted to check with you if I should merge again after adding test api key in .env file. Thank you.

Issue #521, opened 2023-03-05
Update front-end to prepare for Stripe update


DietBepis1:
Amplify has a `Constituents` table to store user data; however, it is currently unused. You can see in the wiki that the `Constituents` table should have a `first_name` and `last_name` field. To prepare the app to store constituent data, the front-end will need to be modified in a number of ways.

**Todos:**
- [ ] Update the `SignName` component to have a first name and last name field. There is already a data validation for name that can be copied over.
- [ ] Update the Vuex store to have a first name and last name property for the `userData` object.
- [ ] There is actually a bug in the `TakeAction.nextPage`method that is overriding userData with an empty object. Fix this method so it doesn't write over the data that the user just filled out.
- [ ] I believe that Lob only has a `name` property for the letter, so create a computed property that concatenates the user's name together for both the letter display and also when the user is passed to `/createLetter`.
- [ ]  Any other spots that need fixing that I might have missed.

lokeshkumar0001:
Hi @DietBepis1, I am a full stack developer but new to open source. Pls guide me on how can I get started with contributing to this organization

paramsiddharth:
@lokeshkumar0001 Hey Lokesh, welcome to open-source and Program Equity! 😊 I would recommend you have a look at the codebase and the issue details and ask any specific questions you come across.

lokeshkumar0001:
@paramsiddharth Thanks for your help. May I know where to ask any specific questions. Do you have any slack or discord channels? or anything like that?

paramsiddharth:
> @paramsiddharth Thanks for your help. May I know where to ask any specific questions. Do you have any slack or discord channels? or anything like that?

@lokeshkumar0001 Sure! Here's our Slack: https://join.slack.com/t/programequity/shared_invite/zt-1p1buco3b-yUXjDaKSz3OSJfaHj88Y4g

smgraywood:
Hi all! I'm going to pick this up, if that's ok with everyone! Glenn sent it to my pod leader as something that I might be able to contribute on.

christina-ml:
I worked on the first todo item for the `SignName` component:
Update the SignName component to have a first name and last name field. There is already a data validation for name that can be copied over.

Notes:
- Remove `name` text field
- Add `firstName` text field
  - Update rule to require `firstName` field
  - Update placeholder text
- Add `lastName` text field
  - Update rule to require `lastName` field
  - Update placeholder text
- Update data object with `firstName` and `lastName` fields
[...]

PR #547, opened 2023-04-04
Update SignName.vue first name & last name fields

christina-ml:
Update the `SignName` component to have a first name and last name field.
For issue #521 Update front-end to prepare for Stripe update

### DIFF
--- a/src/components/SignName.vue
+++ b/src/components/SignName.vue
@@ -3,12 +3,21 @@
     <v-card ref="form">
       <v-card-text>
         <v-text-field
-          ref="name"
-          v-model="name"
-          :rules="[() => !!name || 'This field is required']"
+          ref="firstName"
+          v-model="firstName"
+          :rules="[() => !!firstName || 'This field is required']"
           :error-messages="errorMessages"
-          label="Full Name"
-          placeholder="John Doe"
+          label="First Name"
+          placeholder="John"
+          required
+        />
+        <v-text-field
+          ref="lastName"
+          v-model="lastName"
+          :rules="[() => !!lastName || 'This field is required']"
+          :error-messages="errorMessages"
+          label="Last Name"
+          placeholder="Doe"
           required
         />
         <v-text-field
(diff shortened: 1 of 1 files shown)


paramsiddharth:
image
image

paramsiddharth (changes requested):
Please view the screenshots above. Thank you! :)

src/components/SignName.vue
```diff
@@ -103,7 +112,8 @@ export default {
     name: 'SignName',
     data: () => ({
       errorMessages: '',
-      name: null,
+      firstName: null,
```
DietBepis1, line 115: Technically this works, but I think we should prefer an empty string for uninitialized string fields. That will still be a falsy value and it prevents certain exceptions where a function might expect a string.

src/components/SignName.vue
```diff
@@ -119,7 +129,8 @@ export default {
     computed: {
       form () {
         return {
-          name: this.name,
+          firstName: this.first_name,
```
DietBepis1, line 132: This should be `firstName: this.firstName` and `lastName: this.lastName` since those are what the Vue instance variables are called in `data`

src/components/SignName.vue
```diff
@@ -138,7 +149,7 @@ export default {
 
     methods: {
         addressCheck () {
-          this.errorMessages = this.address && !this.name
+          this.errorMessages = this.address && !this.first_name && !this.last_name
```
DietBepis1, line 152: Same thing here about the `firstName` and `lastName` values.

DietBepis1 (changes requested):

src/components/SignName.vue
christina-ml: Updated uninitialized string fields as empty strings

src/components/SignName.vue
christina-ml: Updated Vue instance variables to be the same as they are called in data

src/components/SignName.vue
christina-ml: Updated firstName and lastName values to match Vue instance variables as they are called in data

christina-ml:
Updated `SignName` component with requested changes.

christina-ml:
Updated SignName component with requested changes. Formatted files with prettier.

Issue #606, opened 2023-10-13
[Documentation] Tests for Cicero API


Alex-is-Gonzalez:
**Files are [here](https://github.com/ProgramEquity/amplify/blob/main/server/__tests__/integration/representatives.test.js)**

**Pre- work:** 
- highlight code snippets and ask Copilot Chat: "Explain this test to me", or use the `/explain` slash command
- Ask Copilot Chat questions about the test framework: 
  - "why do we use the jest framework?"
  - "what are the advantages of jest over chai?"

**Tasks for completion:** 
- [ ] enhance the inline comments: highlight functions and prompt Copilot: "refactor to be more readable with inline comments" 
- [ ] add to the wiki: _highlight the whole file_ and ask Copilot Chat to help generate wiki documentation (ex: "help me generate documentation about this for a wiki"), add to or enhance https://github.com/ProgramEquity/amplify/wiki/Cicero-API-Integration

kalahasv:
Hi, could I work on this?

PR #627, opened 2023-10-24
[Documentation] Tests for Cicero API #606

ktrudickm:
Refactored representatives.test to be more readable with inline comments.


Alex-is-Gonzalez (approved):
Great work

PR #629, opened 2023-10-24
Markdown file detailing the Strip Checkout Implementation steps

EmaleeSoto:
-Created Checkout markdown file to provide documenatation on the Stripe API and how it is used in Amplify
-File includes implementation steps for Stripe API end points


DietBepis1 (changes requested):
This is pretty nice, but I think it would be better to add it into the Amplify wiki.

https://github.com/ProgramEquity/amplify/wiki

CHECKOUT.md
```diff
@@ -0,0 +1,37 @@
+# Documentation - Stripe Functions
+
+## Introduction
+
+Amplify uses Stripe to process donations for non-profit organizations. Use these API endpoints to store and complete transactions.
```
Alex-is-Gonzalez, line 5: Amplify uses Stripe to process donations for non-profit organizations.  Amplify uses these API endpoints to store and complete transactions.

Alex-is-Gonzalez (commented):
There is just a small Typo in the start of the MD file. Once that is changed, I will merge it

Alex-is-Gonzalez (approved):
Thank you for creating the checkout.md

PR #645, opened 2023-10-30
Create Admin Table, Issue #600

### LINKED ISSUE #600: Create Admins Table
### Description
A key thing that the campaign control panel feature introduces is the idea of an authenticated user, i.e. someone who logs in with credentials and has privileges that a casual app visitor lacks. This issue is for creating an `admins` table.

### Spec
The data structure should be something like
| field_name | type  |
|----------------|--------|
| first_name | **string** not null |
| last_name | **string** not null |
| user_name (or email?) | **string** not null |
| password | encrypted **string** not null |
| active | **boolean** not null |
| last_login_at | Date |

What other fields do _you_ think should be added to make this a flexible data structure?

**TIP: You can use `knex` to auto-create an id and timestamps with**
```
table.increments('id')  https://knexjs.org/guide/schema-builder.html#increments
table.timestamps()      https://knexjs.org/guide/schema-builder.html#timestamps
```

### Copilot Prompts
`Explain how to create a Postgres column for password data?`
`Create a knex migration that adds a table called 'admins'`
`Explain a how to an ORM works`

ktrudickm:
Created admin table files. Created seed files to seed entries, created table file for table schema, and created admin model file. Ran commands for db table creation and seeding successfully. Fix #600

Co-authored-by: Eric Wang <ericwang42@outlook.com>

### DIFF
--- a/server/db/seeds/development/seed-admins-table.js
+++ b/server/db/seeds/development/seed-admins-table.js
@@ -1,0 +1,22 @@
+module.exports = {
+  async seed(knex) {
+    // Deletes all existing entries
+    await knex('admins').del()
+
+    // Insert seed entries
+    await knex('admins').insert([
+      {
+        first_name: 'John',
+        last_name: 'Doe',
+        email: 'johndoe@gmail.com',
+        password: 'password'
+      },
+      {
+        first_name: 'Jane',
+        last_name: 'Doe',
+        email: 'janedoe@gmail.com',
+        password: 'password'
+      }
+    ])
+  }
+}
--- a/server/db/seeds/development/seed-admins-table.js
+++ b/server/db/seeds/development/seed-admins-table.js
@@ -1,7 +1,11 @@
+const { encryptPassword } = require('../../../lib/encrypt')
+
 module.exports = {
   async seed(knex) {
     // Deletes all existing entries
     await knex('admins').del()
+
+    const password = await encryptPassword('password')
 
     // Insert seed entries
     await knex('admins').insert([
@@ -9,13 +13,13 @@ module.exports = {
         first_name: 'John',
         last_name: 'Doe',
         email: 'johndoe@gmail.com',
-        password: 'password'
+        password
       },
       {
         first_name: 'Jane',
         last_name: 'Doe',
         email: 'janedoe@gmail.com',
-        password: 'password'
+        password
       }
     ])
   }
--- a/knexfile.js
+++ b/knexfile.js
@@ -1,3 +1,4 @@
+require('dotenv').config()
 const { getEnv } = require('./server/db/util')
 
 const targetEnv = getEnv()
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@vue-stripe/vue-stripe": "^4.4.4",
         "axios": "^0.27.2",
         "axios-cache-interceptor": "^0.10.7",
+        "bcrypt": "^5.1.1",
         "cli-plugin-eslint": "^0.1.0",
         "connect-history-api-fallback": "^2.0.0",
         "dotenv": "^16.0.3",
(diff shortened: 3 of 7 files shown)


server/db/migrations/create-admins-table.js
```diff
@@ -0,0 +1,27 @@
+const tableName = 'admins'
```
DietBepis1, line 1: You should generate this file automatically with `npx knex migrate:make create-admins-table`. This will generate the migration with a timestamp like the other files. You can just copy and paste the contents into the new migration.

That command is not referenced anywhere in the project so don't worry about not having known about it 😄

server/db/migrations/create-admins-table.js
```diff
@@ -0,0 +1,27 @@
...
+  async up(knex) {
+    await knex.schema.createTable(tableName, (table) => {
+      // Auto-incrementing non-nullable unsigned integer primary key "id" field
+      table.increments()
+
+      // Simple fields
+      table.string('first_name').notNullable()
+      table.string('last_name').notNullable()
+      table.string('email').notNullable()
+      table.string('password').notNullable()
```
DietBepis1, line 13: This will create a password column with the type `varchar`. I believe that we want to create the column with a `text` type, so we will need `table.text('password').notNullable()`. That will prevent possible issues with encryption length later on.

server/db/seeds/development/seed-admins-table.js
```diff
@@ -0,0 +1,32 @@
...
+    await knex('admins').del()
+
+    // Insert seed entries
+    await knex('admins').insert([
+      {
+        id: 1,
+        first_name: 'John',
+        last_name: 'Doe',
+        email: 'johndoe@gmail.com',
+        password:
```
DietBepis1, line 13: For the purpose of development, you might want to put something more generic in the seed, like `password`.

DietBepis1 (changes requested):

DietBepis1:
You will need to rebase this PR to resolve that issue metrics merge conflict and clean up the issue metrics commits.

ktrudickm (commented):
Changes reviewed. Ready for a re-review

ktrudickm (commented):
Changes made and ready for re-review

.github/workflows/issue-metrics.yml
```diff
@@ -2,13 +2,52 @@ name: Measure issue timestamp
 
 on:
   issues:
+<<<<<<< HEAD
```
DietBepis1, line 5: Once you make the other changes I have mentioned, hit this PR with one more rebase and I think we will be rid of the issue metrics stuff 🙂

server/db/migrations/20231109224923_create-admins-table.js
```diff
@@ -0,0 +1,27 @@
...
+      table.increments()
+
+      // Simple fields
+      table.string('first_name').notNullable()
+      table.string('last_name').notNullable()
+      table.string('email').notNullable()
+      table.text('password').notNullable()
+      table.boolean('active').defaultTo(true)
+      // Creates an 'created_at' field and 'updated_at' field
+      table.timestamps(true, true)
```
DietBepis1, line 16: I think we should set this to use `table.timestamps(false, true)` instead. That will use Postgres' datetime type, which is a little more flexible.

server/db/seeds/development/seed-admins-table.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  async seed(knex) {
+    // Deletes all existing entries
+    await knex('admins').del()
+
+    // Insert seed entries
+    await knex('admins').insert([
+      {
+        id: 1,
```
DietBepis1, line 9: For `id`, `active`, `created_at`, `updated_at`, let's omit those values from the seed since they are auto-generated. On my local machine, when I pulled this code to test it out, setting the id on the seed de-synchronizes the id counter on the table so when I try to insert a new admin, it tries to use the value `id: 1` and has a problem.

server/utilities/encrypt.js
```diff
@@ -0,0 +1,15 @@
+const bcrypt = require('bcrypt')
```
DietBepis1, line 1: Looks good, but let's move this to a directory called `server/lib/` to make it more obvious that it is coming from a 3rd party.

DietBepis1 (changes requested):

.github/workflows/finish-hackpod.yml
```diff
@@ -10,13 +10,13 @@ jobs:
     runs-on: ubuntu-latest
```
DietBepis1, line 11: Can you remove all these extra files/directories 🙂 :
- .github/workflows
- roiScript

Everything else should be good then!

DietBepis1 (changes requested):

DietBepis1 (approved):

Issue #651, opened 2023-11-01
[Documentation] Update ReadMe graphics to reference "Cicero API"


kazvee:
Branching off of documentation issue #502, the graphics in the [ReadMe](https://github.com/ProgramEquity/amplify/blob/main/README.md) reference "Google Civic API", which is no longer being used. Since "Cicero API" is now used for legislative district matching, the graphics need to be updated.

PR #659, opened 2023-11-01
[Documentation] Update ReadMe graphics to reference Cicero API #651

kazvee:
Update ReadMe graphics to reference "Cicero API" for legislative district matching, instead of "Google Civic API" which is no longer being used.


kazvee:
New PR (https://github.com/ProgramEquity/amplify/pull/750) has been raised for this issue.

Issue #615, opened 2023-10-14
Create Campaigns Rest Endpoints


DietBepis1:
### Description
For the new control panel feature, we will need a few more endpoints to handle `campaign` creation and updating. We currently have a way to get campaigns via API, but we will create new versions of those endpoints under the `v1` api schema. These endpoints should consume JSON.

You may be wondering: _why bother rewriting some code that already exists?_ The answer is because we are adding new features (authentication) and revamping our testing, so I want to start from a clean slate. This is also a pattern you will encounter in the workplace as you do product development so it is good to be able to differentiate legacy functionality from new features.

### Spec
We need to create:

- [ ] **GET** /v1/campaigns; Gets an array of all campaigns in the current context
- [ ] **GET** /v1/campaigns/:id; Gets a single campaign with a given id
- [ ] **POST** /v1/campaigns; Given a single campaign or array of campaigns, create database entries
- [ ] **PUT** /v1/campaigns/:id; Updates the specified campaign when given a JSON payload of modified fields
- [ ] **DELETE** /v1/campaigns/:id; Deletes the specified campaign

**Each of these endpoints should also have appropriate testing!**

### Copilot Prompts
`Help me write a(n) <insert http verb> endpoint using ExpressJS.`

`Explain how Express routes requests.`

`How can I group express endpoints with a version?`

ht-l1:
Hi Glenn, I wanted to work on this but I cannot assign it to myself (perhaps no access to the repo yet?)

briansegura15:
Hey Glenn, gonna get after this one 💪

DietBepis1:
@ht-l1 and @briansegura15 (and whoever reads this later), I would like you to each choose one of the endpoints to build and write tests for. That way you can focus on one digestible piece of work and we can keep PRs small.

danchan020:
Hey all, I'll take the post request: POST /v1/campaigns;

ht-l1:
Can I grab the first one? GET /v1/campaigns;

CodingLady22:
Can I take the second get? GET /v1/campaigns/:id;

briansegura15:
> Can I take the second get? GET /v1/campaigns/:id;

Hey, I already started working on it. The 4th and 5th endpoints mentioned  above are still available.

CodingLady22:
[...]

PR #695, opened 2023-11-03
Add new v1 GET endpoint for /v1/campaign/:id

briansegura15:
This PR for issue #615 adds the new endpoint for retrieving a specific campaign by its ID. 

The endpoint responds with campaign data if found, or a 404 code if the campaign does not exist. Additional wiring between the api folder and api.js file was established. 

Attempts at creating successful tests were made but did not end successfully.


briansegura15:
It seems I somehow managed to make the same commit twice.? Sorry about that.

server/__tests__/integration/v1.campaigns.test.js
```diff
@@ -0,0 +1,36 @@
+// below are where the tests for the campaign endpoints will go, I kept getting errors back and couldnt set this up correctly 😞, i got thrown off by the section about inserting test data into the database
```
DietBepis1, line 1: Don't worry! We can look at your testing issues during Pair Hours.

server/routes/api/v1Campaigns.js
```diff
@@ -0,0 +1,21 @@
+const express = require('express')
```
DietBepis1, line 1: I would like to propose a small change to the way these new files are organized, which is to add some more subfolders. I created an example PR here for reference:

https://github.com/ProgramEquity/amplify/pull/700/files

server/routes/api/v1Campaigns.js
```diff
@@ -0,0 +1,21 @@
...
+const router = express.Router()
+const Campaign = require('../../db/models/campaign')
+
+router.get('/:id', async (req, res) => {
+  try {
+    const campaign = await Campaign.query().findById(req.params.id)
+    if (campaign) {
+      res.send(campaign)
```
DietBepis1, line 9: This is a small nitpick, but I would prefer that we use the following pattern for responses: `return res.status(<status code>).json({ campaign }).end()`. That way we are explicitly adding each piece of the response. Returning the response will prevent code in the route from running after Express sends the response. The `end()` will close the http response which prevents hanging requests in some cases.

DietBepis1 (changes requested):

server/__tests__/integration/v1.campaigns.test.js
briansegura15: Sounds good. I missed yesterdays standup so hopefully we can take a look at it at the next Pair Hours.

server/routes/api/v1Campaigns.js
briansegura15: I will checkout the example PR with the proposed file organization and include it in the new commit.

server/routes/api/v1Campaigns.js
briansegura15: Thanks for the guidance. Going to make the responses more explicit and will implement `return` and `end()` for better request handling.

Issue #617, opened 2023-10-17
User Sessions with Passport.js


DietBepis1:
### Description
To provide some security around our new control panel, we need to create a way in which `admin` users can authenticate themselves so they can change up campaigns. There are many ways to do this and it is important to remember that development must always maintain a balance of security and complexity. Because Amplify is essentially a start-up app, we want to implement something relatively simple, that's also tried-and-true. User sessions fits that bill, while Passport.js is a framework that allows us to change or expand what we do in the future.

Security can be confusing! It's also important and I encourage everyone to get in the weeds with me as we implement this. Here are some resources to start:
- https://www.youtube.com/watch?v=F-sFp_AvHc
- https://github.com/passport/todos-express-password\
- https://www.passportjs.org/tutorials/
- https://www.npmjs.com/package/express-session

### Spec
This is a pretty large endeavor. For the purpose of this issue, the focus should be on implementing the express middleware following the pattern in the above tutorials. One implementation detail that _may_ be different will be how we encrypt passwords (and maybe also usernames in #600).

Testing this will also be a bit more challenging because we will have to walk a narrow line between _testing our implemented functionality_ vs. _testing the library_; this is a common code smell that can be hard to fix. Focus on creating small testable functions that incorporate the code provided by Passport.js.

### Copilot Prompts
`Explain middleware.`
`Show me how to use the express-sessions package.`
`How can I chain Express middleware together?`

KonnyGuo:
Hi I would like to work on this (no permission to assign myself to the issue yet)
I do not see a middleware folder anymore so would a middleware folder need to be made like in amplify/src/middleware

briansegura15:
hey @KonnyGuo , let's team up and tackle this one together.

KonnyGuo:
sure @briansegura15

KonnyGuo:
@DietBepis1 sorry I got some questions. Since middleware is for backend I decided to create a middleware folder in amplify/server/. The middleware code should check if the user is an admin and give them the ability to authenticate themselves. My questions are
1. Where does the code for passport strategies and express-session belong? (I was thinking amplify/server/server.js but I would like to hear what you had plan originally for this.
2. Where is the model containing admin/user in the db?

PR #696, opened 2023-11-03
Implement Passport.js Local Strategy

briansegura15:
This commit for #617 integrates Passport's Local Strategy for user authentication. 
The strategy has been set up to use email as the username field. If the email is not found or the password is incorrect, error messages are returned. 
Was not sure how to implement tests on Users because users do not exist yet.
Co-authored-by: KonnyGuo KonnyGuo@users.noreply.github.com


.github/workflows/issue-metrics.yml
```diff
@@ -8,35 +8,35 @@ on:
   pull_request:
     types: [opened]
     branches:
-      - main 
+      - main
```
DietBepis1, line 11: The oldest comment in the book: Rebase to remove these 😄

server/auth/config/passport.js
```diff
@@ -0,0 +1,43 @@
...
+module.exports = function (passport) {
+  passport.use(
+    new LocalStrategy(
+      {
+        // by default, local strategy uses username, we will override with email
+        usernameField: 'email'
+      },
+      async (email, password, done) => {
+        try {
+          const user = await User.findOne({
```
DietBepis1, line 14: We will have to update this config to use `Admin`s. For password validation, check out `server/lib/encrypt` after you rebase this. That will have the methods you need.

server/auth/config/passport.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+    )
+  )
+
+  passport.serializeUser((user, done) => {
+    done(null, user.id)
+  })
+  passport.deserializeUser(async (id, done) => {
+    try {
+      const user = await User.findByPk(id)
```
DietBepis1, line 37: I think you can just use `Admin.query().findById()` now.
https://vincit.github.io/objection.js/api/query-builder/find-methods.html#findbyid

DietBepis1 (changes requested):
How should we handle merging this story, #617 , and https://github.com/ProgramEquity/amplify/pull/644?

Or should we combine them into one new branch so we can test the whole thing at once? That feels like the right approach to me personally, but I would like your thoughts.

.github/workflows/issue-metrics.yml
briansegura15: Wait was it literally the oldest comments in the book? lol! Rebased ✅

briansegura15: Wait were they literally the oldest comments in the codebase? lol! Rebased ✅

server/auth/config/passport.js
briansegura15: Now that the `Admin` model exists and has been implemented, the mock User model has been removed and config updated. Password validation now being handled by the code in `server/lib/encrypt`

server/auth/config/passport.js
briansegura15: Cleaned up this code to use `.query().findById(id)` instead of `.findByPk(id)`

briansegura15 (commented):
All changes addressed 👍

briansegura15:
As for your first question, I agree that we should merge these couple branches into 1 since it originally was 1 Issue, but we broke it down with Konny to work on it separately.

PR #751, opened 2023-11-12
Created POST endpoint for Admin

jayaruhbee:
POST endpoint for Admin with placeholder Admin paths

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,7 @@
 
 # API key for the Cicero API
 # Check with the team leads for the latest Cicero API key to be used
-CICERO_API_KEY=
+CICERO_API_KEY= 
 
 # Production environment API key for the Lob API
 # This value can be found in Heroku application config var of the same name
--- a/server/api.js
+++ b/server/api.js
@@ -12,7 +12,9 @@ const lob = require('./routes/api/lob')
 const checkout = require('./routes/api/checkout')
 //const twilio = require('./routes/api/twilio')
 const eventLogger = require('./routes/api/event_logger')
-
+// ADD V1 AND ADMIN PATH
+const v1 = require('./routes/api/v1/v1')
+const admin = require('./routes/api/admin/')
 // Created a nested router
 const apiRouter = express.Router()
 
@@ -35,5 +37,8 @@ apiRouter.use('/lob', lob)
 apiRouter.use('/checkout', checkout)
 //apiRouter.use('/twilio', twilio)
 apiRouter.use('/event_logger', eventLogger)
+// ADD V1 AND ADMIN ROUTES
+apiRouter.use('/v1', v1)  
+apiRouter.use('/admin', admin)
 
 module.exports = apiRouter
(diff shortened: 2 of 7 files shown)


.env.example
```diff
@@ -4,7 +4,7 @@
 
 # API key for the Cicero API
 # Check with the team leads for the latest Cicero API key to be used
-CICERO_API_KEY=
+CICERO_API_KEY=
```
DietBepis1, line 7: This is a change that can be reverted since it doesn't have to do with the work in question. It may seem pedantic, but it is better to keep PRs as small as possible.

server/api.js
```diff
@@ -35,5 +37,8 @@ apiRouter.use('/lob', lob)
 apiRouter.use('/checkout', checkout)
 //apiRouter.use('/twilio', twilio)
 apiRouter.use('/event_logger', eventLogger)
+// ADD V1 AND ADMIN ROUTES
```
DietBepis1, line 40: This will work, but it will create two url paths: `api/v1/admin` and `api/admin`. We only want to be able to modify admins under the v1 schema, so the other is unncessary.

server/routes/api/admin.js
```diff
@@ -0,0 +1,29 @@
+const express = require('express');
+const router = express.Router();
+
+// PLACEHOLDER ADMIN
+const Admin = require('./path/to/Admin');
```
DietBepis1, line 5: If you're waiting for another piece of code to be completed for something to work, it is nice (in my opinion) to leave a TODO comment or a note for other developers about it.

server/routes/api/v1/v1.js
```diff
@@ -0,0 +1,18 @@
+const express = require('express')
+const routeCollector = express.Router() 
+
+// Require the actual route files
+const test = require('../../../__tests__/unit/post_admin_test');
```
DietBepis1, line 5: We should not need to require any tests in this way. Can you explain the purpose of creating a test endpoint here?

src/router/index.js
```diff
@@ -13,7 +13,7 @@ const routes = [
     path: '/',
     name: 'Home',
     component: Home
-  },
+  },
```
DietBepis1, line 16: No need for this change.

server/routes/api/admin.js
```diff
@@ -0,0 +1,29 @@
...
+            email: req.body.email,
+            password: req.body.password,
+            active: req.body.active,
+            created_at: req.body.created_at,
+            updated_at: req.body.updated_at,
+        }
+
+        const newAdmin = await Admin.query().insert(admin_data);
+
+        return res.status(201).json({ admin: newAdmin }).end();
```
DietBepis1, line 21: Nice, the correct status code 👍🏽

DietBepis1 (changes requested):

DietBepis1:
@jayaruhbee Can you please rebase and resolve the merge conflicts? I merged a different PR that adds the v1 structure so there are a few minor modifications to make now.

PR #756, opened 2023-11-13
Campaign delete endpoint

CodingLady22:
Add delete endpoint for the campaigns and test using a mock test


server/routes/api/campaigns.js
```diff
@@ -1,6 +1,7 @@
 const express = require('express')
 const router = express.Router()
 const Campaign = require('../../db/models/campaign')
+// const Campaign = require('../../../db/models/campaign')
```
DietBepis1, line 4: Please delete unused imports.

server/routes/api/campaigns.js
```diff
@@ -25,4 +26,16 @@ router.get('/', async (req, res) => {
   }
 })
 
+// Delete endpoint
```
DietBepis1, line 29: This route looks pretty good, but it should be in the v1 campaign router instance 🙂

server/routes/api/campaigns.js
```diff
@@ -25,4 +26,16 @@ router.get('/', async (req, res) => {
...
+router.delete('/:id', async (req, res) => {
+  const id = req.params.id
+  try {
+    await Campaign.query().deleteById(id)
+    res.send({ message: 'Campaign deleted successfully' })
```
DietBepis1, line 34: We should use a pattern of explicitly sending a status code and defining the content-type we are sending back to the client. That will make it easier to understand what is happening in the code and enforce data contracts. You can tweak both `res.send()` calls to:
`res.status(200).json({ message: 'Campaign deleted successfully' }).end()`
`res.status(500).json({ error: error.message }).end()`

server/routes/api/v1/v1.js
```diff
@@ -0,0 +1,13 @@
+const express = require('express')
+const campaigns = require('./campaigns') // If the routes file is named index.js, we can omit it when we do the import instead of writing './campaigns/index'.
+// const { exampleMiddleware } = require('../../../middleware/example')
```
DietBepis1, line 3: Can you please delete the extra comments and any commented out code?

server/__tests__/unit/campaigns.test.js
```diff
@@ -0,0 +1,29 @@
+// server/__tests__/api/campaigns.test.js
+const request = require('supertest')
+const app = require('../../app')
+const Campaign = require('../../db/models/campaign')
+
+jest.mock('../../db/models/campaign', () => {
+  return {
```
DietBepis1, line 7: Nice 😎

server/__tests__/unit/campaigns.test.js
```diff
@@ -0,0 +1,29 @@
...
+    query: jest.fn().mockReturnThis(),
+    findById: jest.fn().mockResolvedValue({ id: 1, name: 'Campaign 1' }),
+    deleteById: jest.fn().mockResolvedValue(true)
+  }
+})
+
+describe('DELETE /api/campaigns/:id', () => {
```
DietBepis1, line 14: Since you have to move the delete endpoint to /api/v1/campaigns as mentioned in a different comment, don't forget to update the tests!

DietBepis1 (changes requested):
Good effort! 😄  Just a few small changes are needed.

server/routes/api/v1/v1.js
CodingLady22: Yes, I have now taken them out

server/routes/api/campaigns.js
CodingLady22: I have deleted all unused imports

server/routes/api/campaigns.js
CodingLady22: The delete route has been transferred to the v1/campaigns folder

server/routes/api/campaigns.js
CodingLady22: I corrected the endpoint to send the appropriate status codes and also included the content-type to be sent back to the client

server/__tests__/unit/campaigns.test.js
CodingLady22: Thank you for the review. I have made and pushed all appropriate changes as you mentioned.

PR #781, opened 2023-11-15
Update issue-metrics.mjs

Dunridge:


Dunridge:
Time from assignment to PR for #781: 42745 ms

PR #800, opened 2023-11-17
[Documentation] Update ReadMe graphics to reference Cicero API #651

kazvee:
Update ReadMe graphics to reference "Cicero API" for legislative district matching, instead of "Google Civic API" which is no longer being used.


Alex-is-Gonzalez (approved):
Thank you for updating the API Graphic

Issue #624, opened 2023-10-24
[Architecture] Welcome messages Action


eprice555:

lingeorge88:
I would like to work on this issue for tomorrow's pair programming hour, thank you!

lingeorge88:
Hi! I tried following the video tutorial to utilize Microsoft graph to see if I can create an event in the user's calendar but my current action fails with the following message: Run anoopt/ms-graph-create-event-msal@v1
⌛ Getting access token...
✅ Got access token
⌛ Getting Graph client...
✅ Got Graph client
⌛ Creating event...
Error: 🚨 Error in createEvent function.
Error: Error
Error: Error

Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-[11](https://github.com/lingeorge88/Password-generator/actions/runs/6700564447/job/18206698846#step:3:12)-github-actions-deprecating-save-state-and-set-output-commands/

this is my current yml setup:
/Users/GeorgeBigballer/Desktop/Bootcamp/Password-generator/.github/workflows/msGraphEvent.yml

lingeorge88:
I have registered an application on Azure AD and configured my client id and secrets following the instructions found here: https://github.com/anoopt/ms-graph-create-event-msal

JulieGunawan:
I would like to add a blog in addition to the documentation, to paint a better picture on all these variables required:
https://anoopt.medium.com/github-actions-create-an-event-with-microsoft-graph-using-msal-ee7aa7b49dc7

lingeorge88:
Hello all, please see pull request #834 for this pull request.  

The current working action as outlined in the pull request follows these sequence of steps: 
1) User assigns themselves to an issue
2) The action triggers a workflow run, and gets the following information : Issue title, assigned date, username
3) The action appends a comment automatically with a Google Calendar Link that will have the issue's title, a tentative due date for the issue (3 weeks from assignment)
4) The user then follows the link to create an event in their perspective calendars. 
5) The user is free to modify the due date and when they would like to be reminded
![Screenshot 2023-12-07 at 3 41 10 PM](https://github.com/ProgramEquity/amplify/assets/115956879/68102592-dc24-4943-9bfd-331fb7bf9fea)
[...]

PR #834, opened 2023-12-07
Github action for creating Google Calendar Reminders upon issue assignment

lingeorge88:
This is a pull request for issue #624 , creating a Google Calendar Reminder when the user assigns themselves to an issue


lingeorge88:
Screenshot 2023-12-07 at 3 41 10 PM
Upon issue assignment, the action will append a comment with a link to create a Google Calendar Event for the assigned issue's tentative due date.  The issue's title will be encoded in the URL link and set as the event title by Default as shown in the following example: 
Screenshot 2023-12-07 at 3 43 33 PM
The user is free to adjust their desired dates, times, reminder types, etc.  The due date is tentative and can be adjusted by the user.

PR #897, opened 2024-03-03
Added new docker workflow for better local development

DietBepis1:
This will hopefully replace or be merged with .devcontainer, but want to leave that as is for more experimentation


DietBepis1:
Merging so I can use it to work on other stuff and get the kinks out before the next cohort.

PR #919, opened 2024-05-08
Issue#Action Cats - Sleepy cat

mpa-mxiang:
For this PR:
- Created a YAML file to trigger sleepy cat action related to [this issue](https://github.com/OpenSourceFellows/amplify/issues/914#issue-2286336339)

sleepy-tom


rcmtcristian:
Hey @mpa-mxiang here to see if I can help, so from what I can see, the issue could be caused due to not having enough permissions on the token you created. Just recheck if you gave the right permissions to the token by going to [https://github.com/settings/tokens](https://github.com/settings/tokens), clicking on it, and hitting 'Edit' if it's a fine-grained token. After that, under 'Repository access,' click on 'All repositories.

image

Then permissions and make sure you have theses permissions active

image

let me know if this is the issue and we can go from there

mpa-mxiang:
Hi @rcmtcristian, thank you for your guidance, I regenerated a fine-grained token by giving those permissions to all repos but I still can see three failed action checks...I am wondering if there is anything else I can do to fix it possibly.

image

rcmtcristian:
@mpa-mxiang So sorry I took so long to come up with a proper answer!  I spoke with Glenn and remembered a similar issue we had about a year ago. It turns out a legacy dependency is causing the problem, interrupting the workflow and preventing this action from working after the update.

So, the issue isn’t with your code. We just need to update some parts of the codebase. again sorry for the delay!

PR #922, opened 2024-05-10
Action workflow creates a cat surprise gif 

ggrossvi:
When you open and reopen a pull request you get a surprise cat gif.  Meow!

### DIFF
--- a/.github/workflows/sleepycats.yaml
+++ b/.github/workflows/sleepycats.yaml
@@ -1,0 +1,26 @@
+name: sleepy-cat
+# on key is trigger when pull request is opened or reopened
+on: 
+  pull_request:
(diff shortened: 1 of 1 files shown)


ggrossvi:
@rcmtcristian I am facing the same check failures as https://github.com/OpenSourceFellows/amplify/pull/919 can you give us further guidance on how to resolve.  Thanks!

rcmtcristian:
@ggrossvi  So sorry I took so long to come up with a proper answer! I spoke with Glenn and remembered a similar issue we had about a year ago. It turns out a legacy dependency is causing the problem, interrupting the workflow and preventing this action from working after the update.

So, the issue isn’t with your code. We just need to update some parts of the codebase. Again sorry for the delay!

PR #923, opened 2024-05-11
Action workflow creates a cat surprise gif

amandaguan-ag:
When you open and reopen a pull request you get a surprise cat gif. Meow!


aaron-muti-420 (approved):
This workflow looks greate

PR #930, opened 2024-06-06
modified merge variables in Tuolumne template

davidchapuis:
Modifications in files other than TuolumneLetterLoad.vue are due to Prettier


breton16:
Merged per @manishapriya94 who's in a bit of an access pinch 👍

